PR #3, opened 2017-08-17
Enable coverage analysis on codecov.io

tovrstra:


tovrstra:
@FarnazH This should be safe to merge. The codecov/project does not work yet because there were no coverage reports yet for master. That will be fixed by merging this PR.

PR #8, opened 2017-08-23
Add AppVeyor continuous integration

FarnazH:
There is a problem with `iodata/molden.py` on Windows due to `file.seek()` and `file.tell()` methods. This is because a different newline character is used on Windows, so seek does not take you back to the exact same place (beginning of the previously read line).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,102 @@
+environment:
+  global:
+    # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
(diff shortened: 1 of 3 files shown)


FarnazH:
P.S. This is a problem with only Python 2.7 on Windows apparently because Python 3.5 and 3.6 changed the way `file.seek()` and `file.tell()` work.

tovrstra:
We should probably rewrite the code not to use seek and tell. That is a rather fragile technique anyway.

tovrstra:
P.S. I've updated the webhook to point to Roberto's account on AppVeyor.

tovrstra:
This is the badge for appveyor:

```[Build status](https://ci.appveyor.com/project/theochem-ci-bot/iodata/branch/master)```

When you push this a new commit (e.g. with the badge) it should start testing on appveyor as well.

tovrstra:
We'll fix the seek and tell later. Merging.

PR #10, opened 2017-09-06
added download template

matt-chan:


matt-chan:
Merging in so Derrick can start hacking

PR #16, opened 2018-05-26
update installation instructions

PeaWagon:
Let me know if the format is okay. I left the old instructions there in case they are still relevant.


matt-chan:
Looks great! Thanks Jen!

PR #20, opened 2018-10-28
Updates

FarnazH:
Here are some fixes/improvements across the module. I will be sharing them in small batches so that it's easier to review. Please let me know if sth is not a good change, and I will change it back.


FarnazH:
@matt-chan the coverage failures are not because I added code that is not tested, it's just because I touched some lines which were not tested in the first place. I would be happy to add more tests to improve the coverage, but I think it makes more sense to do that after some more cleaning/refactoring. So, if you are fine with the commits, I think we can skip `codecov` complaints for now. Please let me know what you think.

README.rst
```diff
@@ -25,7 +25,7 @@ Dependencies
 
 The following dependencies will be necessary for IOData to build properly,
 
-* Python >= 2.7, >= 3.5: http://www.python.org/
+* Python >= 3.5: http://www.python.org/
```
matt-chan: Thanks for catching this. We actually need 3.6 and higher because we're using f-strings. 

```suggestion
* Python >= 3.6: http://www.python.org/
```

matt-chan:
Yep Coverage is fine. It's also just a miniscule change (0.04%).

matt-chan:
Looks good. Thanks @FarnazH!

PR #22, opened 2018-10-30
Updates & Cleanups

FarnazH:
I made some more updates to make the language and format of docstrings, imports, etc. consistent; this should make it easier for the users to read/use the package. There were also some missing documents which are fixed. Please let me know if anything needs to be reverted.


iodata/wfn.py
```diff
@@ -211,20 +216,20 @@ def load_wfn(filename: str) -> Dict:
...
     shell_types = type_assignment[mask]
     shell_types = np.array([shell[i] for i in shell_types])
-    assert shell_map.size == shell_types.size == reduced_size
-    nprims = np.ones(reduced_size, int)
+    assert shell_center.size == shell_types.size == reduced_size
+    shell_nprims = np.ones(reduced_size, int)
     con_coeffs = np.ones(reduced_size)
     # build basis set
-    obasis = {"centers": coordinates, "shell_map": shell_map, "nprims": nprims,
+    obasis = {"centers": coordinates, "shell_center": shell_center, "shell_nprims": shell_nprims,
```
matt-chan: same as above

matt-chan:
Nevermind about the rename. I found myself with some free time unexpectedly! I renamed the corresponding functions in gbasis in https://github.com/theochem/gbasis/pull/30. I'll merge this in as soon as that one passes CI.

matt-chan:
There are a good number of pydocstyle/pycodestyle/pylint errors though. Not sure if they're because old (non-compliant) code was touched though. Would it be possible to fix them up anyways? Should just be a quick ctrl+alt+L in pycharm (if you're on windows or linux).

FarnazH:
Thanks for your feedback @matt-chan 
1. I will fix the QA errors soon. 
2. I think I agree with what you are saying regarding the re-naming of variables. We should wait until after the `gbasis` API is fixed. Also, when I was working on `overlap` implementation today, I noticed that I might not have made the best choices for variable names. So, I will remove the 2 commits which renames variables for now.

matt-chan:
Ah okay, yeah that sounds good. Let's see if we can come up with better names when we meet about gbasis?

PR #23, opened 2018-10-31
Density Matrix (1DM & 2DM)

FarnazH:
This PR is still in progress, and is meant for:
1) implementing 2DM with @tczorro.
2) having a better `IOData` interface to retrieve/compute DM for a given spin.
Please feel free to comment.

### DIFF
--- a/iodata/utils.py
+++ b/iodata/utils.py
@@ -186,3 +186,8 @@ def compute_dm_1st_order(coeffs: np.ndarray, occs: np.ndarray) -> np.ndarray:
         First-order reduced density matrix (1DM).
     """
     return np.dot(coeffs * occs, coeffs.T)
+
+
+def compute_dm_2nd_order(coeffs: np.ndarray, occs: np.ndarray) -> np.ndarray:
+    r"""Compute second-order reduced density matrix (2DM)."""
+    raise NotImplementedError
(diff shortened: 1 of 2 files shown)


iodata/utils.py
```diff
@@ -166,3 +166,28 @@ def check_dm(dm: np.ndarray, overlap: np.ndarray, eps: float = 1e-4, occ_max: fl
     if occupations.max() > occ_max + eps:
         raise ValueError('The density matrix has eigenvalues considerably larger than '
                          'max. error=%e' % (occupations.max() - 1))
+
+
+def compute_dm_1st_order(coeffs: np.ndarray, occs: np.ndarray) -> np.ndarray:
```
matt-chan: Is there a reason this isn't just called compute_1dm(...)?

iodata/utils.py
```diff
@@ -166,3 +166,28 @@ def check_dm(dm: np.ndarray, overlap: np.ndarray, eps: float = 1e-4, occ_max: fl
...
+
+    Returns
+    -------
+    out : array_like
+        First-order reduced density matrix (1DM).
+    """
+    return np.dot(coeffs * occs, coeffs.T)
+
+
+def compute_dm_2nd_order(coeffs: np.ndarray, occs: np.ndarray) -> np.ndarray:
```
matt-chan: Same as above

matt-chan:
I like this format of making changes a lot. It might even be more useful than sitting down and doing a meeting. This way we can work with live code?

iodata/utils.py
tovrstra: or `compute_1rdm`

tovrstra:
This should probably go into the Orbital class. See #24 . I'm not sure yet what the best approach is. Need to sleep over it.

PR #35, opened 2019-01-25
Iterate over file formats

### LINKED ISSUE #6: Every file format should become a module with consistent function & attribute names
We should remove all file-format specific information and code in [iodata/iodata.py](../../blob/master/iodata/iodata.py) because it is fragile. In the documentation of the file formats, [doc/tech_ref_file_formats.rst](https://github.com/theochem/horton/blob/master/doc/tech_ref_file_formats.rst), there is an even worse duplication of information, which is hard to maintain.

These issues can be solved by representing each file format by an object with:
- `name` string with a human-readable name of the file format
- `load` method (accepts file or filename) or None
- `dump` method (accepts file or filename) or None
- `pattern` an fnmatch pattern, e.g. to recognize an extension
- `read_always`: list of names that are always read
- `read_optional`: list of names that may be read
- `write_always`: list of names required when writing a file
- `write_optional`: list of names that can be stored when writing a file
- `ao_index_info`: names of arrays whose indexes correspond to AOs, needed for fixing order and sign conventions. e.g. `{'dm_full': (0, 1), 'orbitals': (0,)}`
- `programs`: list of known other programs that can read or write this file format

FarnazH:
Iterate over modules to avoid lots of if/elif/else in `IOData.from_file` & `IOData.to_file` methods. Fixes #6.


FarnazH:
I fixed all the travis erros. There are two left by `namespace` that can be skipped, I believe.
`Name 'load' found in more than one module` & `Name 'dump' found in more than one module`

tovrstra:
This should be solved by placing the file format modules into a subpackage (e.g. `iodata.formats`) and by adding the following namespace config in `.cardboardlint.yml`:

```yaml
- namespace:
    filefilter: ['- */__init__.py', '- */test_*.py', '- *setup.py', '- tools/*',
                 '- */iodata/formats/*.py', '+ *.py']
```

tovrstra:
Leaving it as it is. See theochem/cardboardlint#77

PR #40, opened 2019-01-25
Remove DM Computation

FarnazH:


FarnazH:
The only travis error is the same namespace error of `load` found in more than one module.

PR #45, opened 2019-02-18
Use pytest instead of nosetests

tovrstra:
Relevant changes:
- Update .travis.yml file to use pytest
- Drop our own tmpdir implementation and use built-in pytest fixtures.
- Use `pytest.raises` instead of `assert_raises`.


tovrstra:
remaining errors on travis are unrelated to this PR.

Issue #25, opened 2018-12-08
Support WFX file format


FarnazH:
Make `wfx.py` module and add `def load(filename: str) -> Dict:` function.
Check http://aim.tkgristmill.com/wfxformat.html

PR #48, opened 2019-03-09
Add load_wfx_low() for WFX format #25

fwmeng88:
1. The function `load_wfx_low()`  consists of 7 inner functions.
2. The function `helper_section()` is built to extract the information give the starting and ending strings.
3. The function `helper_str()` is built to extract the string type information, namely title, keywords and model_name.
4. The function `helper_int()` is built to get the int type information with numpy.ndarrays returned.
5. The function `helper_float()` is built to extract the float type information with numpy.ndarrays returned.
6. The function `energy_gradient()` is built to extract the energy gradient.
7. The function `helper_mo()` is built to extract the Molecular Orbital Primitive Coefficients and Molecular Orbital Counts.
8. Add checks for tags, `check_tag()`, keywords and number of perturbations. 
9. Add test data and test codes for `load_wfx_low()`


fwmeng88:
It seems that I put too many statements for `wfx.py`, saying `36:-      iodata/wfx.py  too-many-statements Too many statements (71/50)`. Would you please clarify which variables are required and which one can be skipped over? I think this would be one thing I need to fix in order to pass all the checks. Thank you in advance. @tovrstra @FarnazH 

```
def load_wfx_low(filename: str) -> Tuple:
    # implementation
    return
        title, keywords, model_name, atom_names, num_atoms, num_primitives, \
        num_occ_mo, num_perturbations, num_electrons, num_alpha_electron, \
        num_beta_electron, num_spin_multi, charge, energy, \
        virial_ratio, nuclear_viral, full_viral_ratio, mo_count, \
        atom_numbers, mo_spin_type, coordinates, centers, \
        primitives_types, exponent, mo_occ, mo_energy, gradient_atoms, \
        gradient, mo_coefficients
```
#25

tovrstra:
@fwmeng88 Thanks for this nice contribution. You may be able to solve the `Too many statements (71/50)` problem by moving the helper functions outside the load function, unless they use local variables from define in the body of the `load` function.

If you want to stress that these helper functions are for internal usage of your module only, you can prefix the function names with an `_`. (That is just a small detail.)

I guess, you still intend to add a higher-level load function, but I would not mind this being merged with the low-level load only.

iodata/wfx.py
```diff
@@ -0,0 +1,290 @@
...
+                     'MO Number',
+                     'Energy = T + Vne + Vee + Vnn',
+                     'Virial Ratio (-V/T)']
+    if set(tags_header).intersection(set(tags_required)) != \
+            set(tags_required):
+        diff = set(tags_required) - set(tags_header).intersection(
+            set(tags_required))
+        err_str = ', '.join(diff)
+        err_str += 'are/is required but not present in the WFX file.'
+        raise AssertionError(err_str)
```
tovrstra: `AssertionError` is typically only used for logical consistency checks inside a code, normally something a user of a library should encounter. `ValueError` or `IOError` would be more suitable.

fwmeng88:
Right now, this small module is capable of extracting information from the WFX file, but still not higher-level load function. Some codes are implemented to check:
1. If all required tags are present according to the official documentation of WFX, http://aim.tkgristmill.com/wfxformat.html
2. If all the tag are present pairwise

We might need to format the `AssertionError` as `warnings.warn()` message, which would be a more reasonable choice.

iodata/wfx.py
fwmeng88: Going to fix it soon. Thank you for the update.

fwmeng88: Fixed at 77dca55

tovrstra:
@fwmeng88 Would it be possible to have a formally correct wfx file without the tags in place? I guess this depends a bit on the programs producing WFX files. The code already looks good to me, so I'm fine with merging, unless you want to continue working in this PR. Another possibility is to make a new PR for the following steps. The choice is up to you.

fwmeng88:
> Looks good to me, even without the higher-level code in place, this is already useful and suitable for merging.

Thank you. Will work on the higher-level `load()` function soon.

fwmeng88:
> @fwmeng88 Would it be possible to have a formally correct wfx file without the tags in place? I guess this depends a bit on the programs producing WFX files. The code already looks good to me, so I'm fine with merging, unless you want to continue working in this PR. Another possibility is to make a new PR for the following steps. The choice is up to you.

Shall we merge it first and I do another PR later?

Would you be more specific for the formally correct WFX file without the tags in place? If there are no tags in WFX file, how could we distinguish all the different information. I think I didn't get your point. A sample text might help a lot.

tovrstra:
Just merged.

What I meant is that, if the absence of tags makes the file useless or if it would suggest the file is unreliable, raising an Exception (instead of a warning) is appropriate.

Issue #42, opened 2019-01-27
Intuitive representation of molecular basis set


tovrstra:
Our current representation of molecular Gaussian basis sets is not ideal:

- not very intuitive
- not very pythonic
- mainly geared toward the FCHK format, but it requires non-intuitive manipulations for all other formats.

The Gaussian orbital basis can be described by a single attribute:

`obasis` of the type `namedtuple('OBasisInfo', ['centers', 'shells', 'type', 'conventions'])`

where:

- `centers` is an array with of shape `(ncenter, 3)`
- `shells` is a list of objects of the type `namedtuple('Shell', ['icenter', 'iatom', 'angmoms', 'exponents', 'contractions'])`
- `type`: any of `'cart'` or `'pure'`
- `conventions`: a dictionary with as key an angular momentum character (`'s'`, `'p'`, ...) and as value a list of basis function strings, e.g.
```python
# alphabetically ordered Cartesian functions
['1'],
['x', 'y', 'z'],
['xx', 'xy', 'xz', 'yy', 'yz', 'zz'],
# wikipedia ordered real solid spherical harmonics
['s'],
['pc1', 'pc0', 'ps1],
['dc2', 'dc1', 'dc0', 'ds1', 'ds2'],
```
Any of these strings can be prefixed with a minus sign (`-`) to denote sign conventions.

where:
- `icenter`: an integer referring to a row of `centers`
- `iatom`: an integer for the atom on which the shell is centered, or `None`. This could be convenient when implementing Pulay forces.
- `angmoms`: a string where each character represents an angular momentum of one of the contractions in the shell. The length equals the number of contractions: `len(angmoms)=ncon`. Any of the following can be used: `["s", "p", "d", "f", "g", "h", "i", "k", "l", "m", "n", "o", "q", "r", "t", "u", "v", "w", "x", "y", "z", "a", "b", "c", "e"]`
- `exponents`: an array of exponents of primitives, with shape `(nprim,)`.
- `contractions`: an array with contraction coefficients, with shape `(ncon, nprim)`. These coefficients assume that the primitives are L2-normalized, but contractions are not necessarily normalized.

**Remarks**:
- A main downside of this idea is that IOData objects will become harder to store as `npz` files.
- Attributes `permutation` and `signs` can be removed and should be replaced by convenience functions.

**Example of an iterator over all basis functions:**
```python
def iterate_basis(obasis):
    for shell in obasis.shells:
        for angmom in shell.angmoms:
[...]

Issue #41, opened 2019-01-25
Solidify API & Variable Names


FarnazH:
We should make the necessary changes to vague and long variable names... See issue #1

FarnazH:
#1

tovrstra:
I'll make a list of additional name/API changes w.r.t #1:

**Name changes**
- `numbers` -> `atnums`
- `pseudonombers` -> `atecpnums`
- `coordinates` -> `atcoords`
- `polar` -> `...` (dipole-dipole polarizability matrix)
- `cell` -> `cellvecs` (matrix whose rows or cell vectors, shape `(nvec, 3)`
- `ms2` -> `mult`

**Consolidation of ever-growing number of attributes**
- `mulliken_charges`, `npa_charges`, `esp_charges` -> `atcharges` is a dictionary where the key is a string describing the type of atomic charge and the value is a corresponding array of atomic charges. Keys should be specific and somewhat self-explaining, e.g. `'mulliken'`, `'natural'`, `'lowdin'`, `'hly'`, `'msk'`, etc.
- `one_mo`, `kin`, `na`, `olp` -> `one_body_integrals` is a dictionary where the key is any of `'core'`, `'kin'`, `'na'`, `'olp'`, followed by a suffix `'_ao'` or `'_mo'`
- `two_mo`, `er` -> `two_body_integrals` is a dictionary where the key is any of `'er'`, `'two'`, followed by a suffix `'_ao'` or `'_mo'`
- `dm_*` -> `one_rdm` is a dictionary with various types of 1-particle reduced density matrices. Standard keys are `'scf'`, `'post_scf'`, `'scf_spin'`, `'post_scf_spin'`.

**Things to be removed** because they are not needed at this stage or in the near future
- `links`
- `permutation` and `signs` (See #42)

**New attributes for `IOData`** for generating QC input files, keeping QM/MM in mind)
- `runtype`: `'energy'`, `'opt'`, `'freq'`, ...
- `obasis_name`, e.g. `'6-31g'`, or `'sto-3g'`
- `extcharges`, array with values of external charges, with shape `(nextcharge,)`
- `extcoords`, array with positions of external charges, which shape `(nextcharge, 3)`
- `charge`, the net charge (should not be a new attribute, but something that is coupled to `nelec`.

**New attributes for `IOData`** for reading useful properties out of QC calculations
- `atforces`, forces in Cartesian coordinates, shape `(natom, 3)`
[...]

Issue #24, opened 2018-11-19
Add orbitals helper


matt-chan:
It's too complicated to make an Orbital instance right now. You need to assign the attributes by hand, which is really clunky. ie

```
a = Orbitals(*mol.occ_alpha)
a.coeffs[:] = mol.occ_alpha_coeffs
a.energies[:] = mol.occ_alpha_energies
a.occupations[:] = mol.occ_alpha_occs
```

Instead we should introduce a new parameter, `orb_alpha_params` (or something along those lines) to be fed into Orbitals in the same way we do GBasis (this also requires an API change to Orbitals within meanfield). Then the API becomes something like
```
Orbitals(**mol.occ_alpha_params)
```

FarnazH:
@matt-chan, there is no `Orbital` class in `iodata` at this point, is there? I think this is relevant to the other thread regarding `Orbital` package vs. class.

matt-chan:
Sorry for the delay @FarnazH. No, there's no orbital class right now. But even to instantiate an Orbital instance from meanfield or from another orbital package, it's too clunky I think. We should instead give it all the parameters needed as a dictionary. Right now a developer would need to use the first code-block just to create an orbital instance.

tovrstra:
Re-introducing an Orbital class is certainly an option, but it should remain focussed on facilitating IOData, not too many adding computational features. I would suggest to have one class, with possibly two instances, one for alpha and one for beta orbitals. I still need to figure out how to make it the most convenient for chemtools and denspart.

tovrstra:
One `mo` attribute in `IOData` that is flexible enough to handle different cases (restricted, unrestricted, generalized, open-/closed-shells)

`mo` is an instance of type `namedtuple('MolecularOrbitals', ['type', 'energies', 'occupations', 'irrep', 'coefficients'])`

where

- `type` is any of `'generalized'`, '`restricted`', `'unrestricted'`
- `energies` is an array of orbital energies
- `occupations` is an array with occupation numbers. Normally in the range [0, 1], but can be [0, 2] for restricted
- `irrep` a list of strings
- `coefficients` one array with orbital coefficients, with rows corresponding to basis functions and columbs corresponding to orbitals. The number of rows equals the number of spatial basis functions, except for generalized where it is twice as large (once for alpha and once for beta). The number of columns equals the number of orbitals. In case of `'unrestricted'`, the first half of the columns are alpha and the second half is beta.

tovrstra:
[...]

PR #51, opened 2019-03-16
Add Q-Chem caculation output format support

fwmeng88:
The original code was taken from https://github.com/molmod/tamkin/blob/a15141745c90abbe8dadb1394f538e30a3523d48/tamkin/io/qchem.py#L36-L165. 
1. The code was reformated and add functionality to handle all the important information in the qchem file.
2. Add constants values for amu, avogadro, and calorie for unit conversions.
3. Add tests and the corresponding data.


fwmeng88:
It seems the coverage for is not high enough. Do you know how to resolve it, please? The tests for `qchem.py` is 100% coveraged right now. @tovrstra @evohringer @FarnazH

tovrstra:
@fwmeng88 Don't worry too much about the coverage analysis. The results from Codecov are at times incomprehensible.

Is it ok that I review your PR after IOData has gone through some major API changes? Once we have #42, #41 and #24 fixed, the load functions will become easier to write.

fwmeng88:
> @fwmeng88 Don't worry too much about the coverage analysis. The results from Codecov are at times incomprehensible.
> 
> Is it ok that I review your PR after IOData has gone through some major API changes? Once we have #42, #41 and #24 fixed, the load functions will become easier to write.

No problem. I will wait and try to fix it after the API changes. Thank you.

PR #52, opened 2019-03-17
Use numpy.testing for unit tests

### LINKED ISSUE #32: Use np.testing when appropriate in unit tests
When comparing floating point results, `numpy.testing` offers nicer functions than a bare assert. Many of the unit tests should be converted to use `numpy.testing` instead.

fwmeng88:
Change to `numpy.testing` style for `test_wfn.py`.
More formatting work is coming.

Fixes #32


iodata/test/test_chgcar.py
```diff
@@ -21,56 +21,62 @@
...
-    assert (mol.numbers == [8]).all()
-    assert abs(volume(mol.rvecs) - (10 * angstrom) ** 3) < 1e-10
+    assert_equal(mol.numbers, 8)
+    assert_allclose(volume(mol.rvecs), (10 * angstrom) ** 3, atol=1.e-10)
     ugrid = mol.grid
-    assert len(ugrid['shape']) == 3
-    assert (ugrid['shape'] == 2).all()
-    assert abs(ugrid['grid_rvecs'] - mol.rvecs / 2).max() < 1e-10
+    assert_equal(len(ugrid['shape']), 3)
+    assert_equal(ugrid['shape'], 2)
```
tovrstra: Very minor comment and I'm only mentioning this because it could be useful to know. The above two lines can be written as:
```python
assert_equal(ugrid['shape'], [2, 2, 2])
```

iodata/test/test_chgcar.py
```diff
@@ -21,56 +21,62 @@
...
     d = mol.cube_data
-    assert abs(d[0, 0, 0] - 0.78406017013E+04 / volume(mol.rvecs)) < 1e-10
-    assert abs(d[-1, -1, -1] - 0.10024522914E+04 / volume(mol.rvecs)) < 1e-10
-    assert abs(d[1, 0, 0] - 0.76183317989E+04 / volume(mol.rvecs)) < 1e-10
+    assert_allclose(d[0, 0, 0],
+                    0.78406017013E+04 / volume(mol.rvecs), atol=1.e-10)
+    assert_allclose(
+        d[-1, -1, -1], 0.10024522914E+04 / volume(mol.rvecs), atol=1.e-10)
+    assert_allclose(d[1, 0, 0],
+                    .76183317989E+04 / volume(mol.rvecs), atol=1.e-10)
```
tovrstra: Again very minor. A max line length of 100 characters is ok.
See https://github.com/theochem/iodata/blob/master/.pycodestylerc
Anyway, no need to change anything. We should consider a tool lick black or yapf to fix formatting issues automatically.

iodata/test/test_chgcar.py
fwmeng88: It's good to know for its simplicity. Thank you.

iodata/test/test_chgcar.py
fwmeng88: I was using `autopep8` assuming the max line length be 80. Probably we should use the same formatter to make it consistent, `black` or `yapf`. Will update soon.

PR #53, opened 2019-03-19
Replace IOData methods to_file and from_file by functions

tovrstra:
New functions are ``load_one`` and ``dump_one``. This is the first step described in #7. I'm trying to get this done soon to make life easier for @evohringer and @fwmeng88.

### DIFF
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -273,101 +273,104 @@ class IOData:
         elif hasattr(self, 'pseudo_numbers'):
             return len(self.pseudo_numbers)
 
-    @classmethod
-    def from_file(cls, *filenames: str) -> 'IOData':
-        """Load data from a file.
 
-        This routine uses the extension or prefix of the filename to
-        determine the file format. It returns a dictionary with data loaded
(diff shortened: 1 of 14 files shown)


evohringer:
I think the changes are reasonable, but cannot really aprove it.

tovrstra:
@evohringer Do you mean that you have no approve functionality on the github website? If needed, I can try to fix some settings.

tovrstra:
`dump_one1 has become a function (indented back four spaces compared to old code), so it does not have the `self` argument anymore. Instead it needs something to know which information to write to a file, which has now become the first `iodata` argument of the function. It is very similar to the old method, but no longer tied closely to the `IOData` class.

evohringer:
> `dump_one` has become a function (indented back four spaces compared to old code), so it does not have the `self` argument anymore. Instead it needs something to know which information to write to a file, which has now become the first `iodata` argument of the function. It is very similar to the old method, but no longer tied closely to the `IOData` class.

Thanks. I just saw the indentation.

PR #54, opened 2019-03-20
All load functions get a LineIterator object instead of a filename as argument

tovrstra:
Some additional changes were needed:

- All file format modules were moved to `iodata.formats`. This makes it easy to lint them differently, and it also makes the code easier to follow for contributors.

- The LineIterator was introduced, instead just providing file objects to the load functions in iodata.formats. This was needed for generating helpful error messages and to force the load functions to
  read data from the files with the next function. This will later facilitate reading trajectories.

- Nested functions were unnested in several modules to make pylint happy.

- The FCHK code was rewritten for clarity and to avoid classes, due to many pylint errors.

- CONTRIBUTING.rst, to explain how the new load functions work and how to contribute code.

This is the second step to address #7, arguably the biggest one.


fwmeng88:
Shall we ask the contributors to make `git rebase` to make the git history much cleaner?

tovrstra:
I wouldn't suggest rebasing as the standard method to merge pull requests. I agree it results in a nice linear git history, but I don't find linearity very important. (I used to but I changed my mind after a few years.) My main reason for not suggesting rebase is that it can be a lot harder to resolve conflicts when git rebase than with git merge, i.e. when there are pieces of code touched in both the PR and the master branch. These conflicts are always easier to resolve with a merge than a rebase. There are also conceptual arguments against rebasing, i.e. some find it less nice, but it is mostly a matter of personal preference.

That said, rebasing can often be very handy to keep local branches organized, and for that I also use it a lot. As soon as I've pushed a branch to github (or another public server) to make a pull request, I try not to rebase (or force push) anymore. It makes it harder for reviewers to follow what is going on. The only exception is when a pull request is becoming a total mess. In that case, I try to just close the pull request and make a new one with a cleaned-up branch.

tovrstra:
@FarnazH @evohringer I'm going to merge already so @fwmeng88 can work with the new api for his upcoming PRs. Sorry for the rush.

PR #59, opened 2019-03-23
Change dump API: dump functions now take file object is first arg

tovrstra:


tovrstra:
This addresses one of the points in issue #7.

PR #60, opened 2019-03-26
New MolecularBasis API

### LINKED ISSUE #30: Write Molden files in double precision
The molden format does not specify how floats should be written. We can write orbital energies, occupation numbers and orbital coefficients in double precision scientific notation `"%.15e"`.

### LINKED ISSUE #42: Intuitive representation of molecular basis set

tovrstra:
Fixes #42.
Fixes #30.

Differences w.r.t. the plans dicsussed in #42:

- Class names: `MolecularBasis` instead of `OBasisInfo` and `Shell` instead of `ShellInfo`. The info suffix is somewhat redundant and the new datatype can also describe density basis sets.

- `basis.type` was dropped and was replaced by `Shell.kinds`. The original idea could not describe the common situation that some shells are pure while others are Cartesian. The new form is more general than what one may encounter in the supported file formats, similar to what we had in HORTON2. This is a little overkill, but it makes it easier to write algorithms with the new data structure. E.g. `zip(shell.angmoms, shell.kinds)` is convenient and used in several places.

- `shell.contractions` has become `shell.coeffs`. This was changed after David's comment that contractions can be many things, depending on who you ask. `concoeffs` was found too descriptive. The shape is transposed: `(nprim, ncon)` instead of `(ncon, nprim)` to correspond with the usual rows and columns found in practically all file formats.

- `shell.iatom` was dropped. We need to address this in future, but I'd like to postpone until #41 is fixed and ghost atoms are dealt with more sensibly by the `IOData` class. (An issue should be made for this.) Including it in this PR would have added more clutter, most likely to be removed again later.

- Related to the previous point, another issue should be opened for `MolecularBasis.centers`. It is nearly redundant and would become 100% redundant with #41.

- `shell.angmoms` has become a list or array of integers instead of characters, because integers are much easier to use in the rest of the code.

- `basis.primitive_normalization` to make the distinction between orbital and density basis sets.

Additional changes needed to make everything work:

- A new `iodata.basis` module was added, containing all the new data structures and related functions to convert between different ordering conventions.

- All `load` and `dump` functions for wavefunction files have been updated to use the new structure. It had a rather big impact on these modules, but always resulted in cleaner code. Permutation vectors must no longer be hard-coded.

- The routine for the overlap integrals was rewritten, at least the pure Python part, and became much simpler.

- `iodata.load_one` has become a lot simpler. It now only loads one file at a time. The old API, which allowed multiple related files to be loaded at once, is no longer needed. It used to be formally necessary, e.g. to convert integrals from a gaussian log file into the right order, which could only be done by loading a corresponding FCHK at the same time. Now we keep track of ordering conventions and such files can be loaded separately. (This will make it easier to add new features to `load_one` in future, e.g. to manually specify the file format when needed.)

-  `shell_types`, i.e. angular momenta with a sign convention (negative for pure functions) no longer exists. Instead angular momentum and the kind of basis function (Cartesian or pure) have become separate attributes in the `Shell` class.

- Lots of unit tests were added to increase the coverage. `iodata.overlap` and `iodata.basis` should have 100% coverage.

I know this is a big PR, not ideal, but the intermediate code was even worse because it had to keep both the old and the new API in the air.

### DIFF
--- a/.pycodestylerc
+++ b/.pycodestylerc
@@ -1,3 +1,3 @@
 [pycodestyle]
 max-line-length=100
-ignore=E741
+ignore=E741,W503
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -1,0 +1,306 @@
+# -*- coding: utf-8 -*-
+# IODATA is an input and output module for quantum chemistry.
+#
+# Copyright (C) 2011-2019 The IODATA Development Team
+#
+# This file is part of IODATA.
+#
+# IODATA is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 3
+# of the License, or (at your option) any later version.
+#
+# IODATA is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>
+#
+# --
+"""Utility functions for working with basis sets."""
+
+from functools import wraps
+from numbers import Integral
+from typing import List, Dict, NamedTuple, Tuple
+
+import numpy as np
+
+__all__ = ['angmom_sti', 'angmom_its', 'Shell', 'MolecularBasis',
+           'convert_convention_shell', 'convert_conventions']
+
+ANGMOM_CHARS = 'spdfghiklmnoqrtuvwxyzabce'
+
+
+def _alsolist(f):
+    """Wrap a function to accepts also list as first argument and then return list."""
+    @wraps(f)
+    def wrapper(firsts, *args, **kwargs):
+        if isinstance(firsts, (Integral, str)):
+            return f(firsts, *args, **kwargs)
+        return [f(first, *args, **kwargs) for first in firsts]
+    return wrapper
+
+
+@_alsolist
+def angmom_sti(char: str) -> int:
+    """Convert an angular momentum from string to integer format.
+
+    Parameters
+    ----------
+    char
+        Character representation of angular momentum, (s, p, d, ...)
+
+    Returns
+    -------
+    angmom
+        An integer representation of the angular momentum. A list, if a list of
+        words was given.
+
+    """
+    return ANGMOM_CHARS.index(char.lower())
+
+
+@_alsolist
+def angmom_its(angmom: int) -> Dict:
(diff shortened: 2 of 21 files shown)


tovrstra:
Thanks. I find it also hard to know in advance how every detail should be implemented. This changes is mainly a big move in approximately the right direction, without doing anything obviously stupid. We can still refine things later.

PR #65, opened 2019-04-07
Fix pycodestyle E302 expected 2 blank lines

FarnazH:
Fix pycodestyle E302 expected 2 blank lines


FarnazH:
@tovrstra Codecov report doesn't make sense to me.

PR #68, opened 2019-04-09
Add duplicate-code to global list of disabled pylint checks

tovrstra:
These warnings should not be enforced as errors. These are only sometimes but certainly not always useful suggestions. An additional problem with the code duplication check is that it does not work when pylint is executed in parallel.


tovrstra:
@FarnazH I'm going to start merging small fixes, so I can make place for bigger PRs. I assume this rather harmless but feel free to bring it up again if needed.

PR #69, opened 2019-04-09
Read version from version.py in setup.py

tovrstra:
The old code does not work when setup.py is executed from a source distribution. The new code should still work on Travis and, unlike older implementations, is not sensitive to the type of quotes used for the version string.


tovrstra:
@matt-chan @FarnazH Same here. I'll be merging this in preparation for bigger PRs.

PR #71, opened 2019-04-14
Add more linters and clean up accordingly

tovrstra:
This is a rather boring change. I'll let it open for a few days in case you want to comment.


tovrstra:
Thanks.

PR #75, opened 2019-04-19
XYZ & FCHK Trajectory support

### LINKED ISSUE #7: Add API to load trajectories or databases, one molecule at a time
Typical examples are XYZ files (e.g. for MD trajectories) or SDF files (databases of molecules). For both formats, it is desirable to have an iterator over the repeated items (usually molecules), without reading the entire file at once into memory.

tovrstra:
Fixes #7 

I've extended the IOData API in such a way that also more involved trajectory formats can be implemented. For example, a Gaussian IRC output is not simply a concatenation of frames, all having the same format. Instead this file also contains various amounts of extra data, which need to be ignored carefully. For these more involved formats, a more elaborate `load_many` function can be implemented in a module in `iodata.formats`.


tovrstra:
@FarnazH Thanks. I'll leave this open till somewhere next week in case @evohringer would like to comment.

tovrstra:
Things should still be simplified a bit: `load_many` and `dump_many` can be removed from `xyz.py` until we encounter a format that really needs such functions.

tovrstra:
I've added support for trajectory data from FCHK files (optimization, relaxed scan and IRC). This is an example where the more elaborate API with format-specific `load_many` or `dump_many` functions is needed.

tovrstra:
@FarnazH Ready for review.

tovrstra:
I'm merging this to make room for upcoming pull requests. We need to move on.

PR #82, opened 2019-05-08
Rename cell to cellvecs and related changes

tovrstra:
- Rename grid to ugrid (needs better solution for which I'll open an
  issue).
- Rename also rvecs to cellvecs (another name in use for the same thing)
- Rename grid_rvecs to axes (common terminology for cube file).
- Drop pbc field from ugrid (no longer used).

(This is one of the bullet points in #41.)


tovrstra:
Issue for ugrid: #83

PR #84, opened 2019-05-09
IOData attribute names: atforces, athessian, atfrozen, atmasses

tovrstra:
We should consider using atgradient instead of atforces. The former is more common. The links attribute was removed from the documentation because it is no longer used.


tovrstra:
Step towards #41.

PR #98, opened 2019-05-12
Try using attrs module

### LINKED ISSUE #73: Get rid of `# pylint: disable=no-member`
When this is implemented, we should also find a way to get rid of the `# pylint: disable=no-member` lines in most unit tests. This pylint warning is caused by the rather sloppy design of the `IOData` class. (Attributes are not initialized in `__init__`.)

To solve the `no-member` issue, one should in principle just initialize all attributes in the constructor `__init__`, but this would get very tedious for IOData. Some alternatives are discussed in the documentation of the `attr` package: https://www.attrs.org/en/stable/why.html This overview is mainly intended to advocate the usage of the `attr` package, so critical reading is advised. Still `attr` offers a lot of nice features for the IOData class, including:

- `attr` goes well with type hinting, similar to `typing.NamedTuple`. See https://www.attrs.org/en/stable/types.html
- `attr` supports validators, for which we have now our own rather ugly code. See https://www.attrs.org/en/stable/examples.html#validators
- `attr` makes `__slots__` easy, which could be useful to force all data with non-standard names to be put an `extra` dictionary. (I'll comment on it below.) See https://www.attrs.org/en/stable/glossary.html#term-slotted-classes

### LINKED ISSUE #77: Add convenience properties to `MolecularOrbitals`: `coeffs_a`, ...
It may be useful to make `coeffs_a` and `coeffs_b` properties of the MolecularOrbitals class. Similarly, one could make `energies_a`, `energies_b`, `occs_a` and `occs_b`.

All of these should return None on the generalized case. In case of restricted, `coeffs_b` and `energies_b` should be None, and elements of `occs_a` and `occs_b` should be at most one

_Originally posted by @tovrstra in https://github.com/theochem/iodata/pull/76_

tovrstra:
Fixes #77 
Fixes #73 

The two issues were solved at the same time because #77, implied making different classes for orbitals (to avoid lots of ifs), but subclassing of NamedTuple subclasses do not work, such that I had to switch to attrs to fix this, which was mainly the point of #73.

The fix of #73 is only partially successful. Not all pylint exclusions could be removed. Several bugs were found by using the attrs module, see https://www.attrs.org/en/stable/. Atts is an extra dependency but not a major concern because attrs is also a dependency of pytest, which we use anyway.

Other notable changes:
- This PR is API breaking because the constructor for the orbitals got simplified.
- Load irreps from Molden file (for testing new features mainly).

### DIFF
--- a/iodata/formats/cp2k.py
+++ b/iodata/formats/cp2k.py
@@ -25,7 +25,8 @@ import numpy as np
 from scipy.special import factorialk
 
 from ..basis import angmom_sti, MolecularBasis, Shell, HORTON2_CONVENTIONS
-from ..utils import LineIterator, MolecularOrbitals
+from ..orbitals import MolecularOrbitals
+from ..utils import LineIterator
 
 
 __all__ = []
--- a/iodata/formats/cp2k.py
+++ b/iodata/formats/cp2k.py
@@ -25,7 +25,7 @@ import numpy as np
 from scipy.special import factorialk
 
 from ..basis import angmom_sti, MolecularBasis, Shell, HORTON2_CONVENTIONS
-from ..orbitals import MolecularOrbitals
+from ..orbitals import RestrictedOrbitals, UnrestrictedOrbitals
 from ..utils import LineIterator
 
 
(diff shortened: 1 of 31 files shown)


tovrstra:
too much of a mess...

PR #103, opened 2019-05-14
Q-Chem format

fwmeng88:
1. Add `load_qchem_low` functionality
2. Add Avogadro constant and calorie
3. Add tests 

TODO:
1. Add `load_one`
2. Use `assert_allclose` instead of assert().all
3. Fix pylint


tovrstra:
@fwmeng88 When you want me to review it, just ping me. I'll edit your top post a little to show a trick with check boxes on github.

fwmeng88:
Dear Too, I am adding `load_one` function. But I think I am not sure how to deal with orbtial, basis and mo part. Would you make it more explicit? Such as how to deal with orbitals and what should be included in mo dictionary. Thank you.
@tovrstra

tovrstra:
@fwmeng88 Sorry for the late reply. your message was lost in my inbox. I'll try to clarify a few things here first, because I'm not sure yet how to improve the developer documentation to clarify things.

A few points that might be helpful:

- The `mo` attribute contains all the information on the orbitals, that is orbital energies, orbital coefficients, occupation numbers and irrep labels. If some of these are not present, you can leave the corresponding fields to `None`. To make an `mo` attribute, you can create an instance of `MolecularOrbitals`, which is defined here: https://github.com/theochem/iodata/blob/master/iodata/orbitals.py

- In the output file you included for testing, the basis is not defined in detail, only `BASIS sto-3g`, which can be stored as `basis_name`. You do not have to make an `obasis` object, because the corresponding information is not present in the QChem output.

Does this help? If you have any further questions, please let me know. I'll make a few comments on the code in an instant.

iodata/formats/qchem.py
```diff
@@ -0,0 +1,251 @@
...
+            break
+
+    return title, runtype, basis, exchange, atnames, atnums, num_atoms, num_sym, \
+        masses, atcoords, polar_matrix, athessian, nucl_repul_energy, \
+        num_alpha_electron, num_beta_electron, mol_charges, \
+        mulliken_charges, energy, alpha_mo_energy, beta_mo_energy, \
+        vib_energy, enthalpy, entropy
+
+
+def load_one(lit: LineIterator, lit_hess: LineIterator = None) -> dict:
```
tovrstra: All `load_one` functions in `IOData.formats` should follow the same API, i.e. just take one `lit` argument. In this case, I would suggest to make a second module `iodata.formats.qchem_hessian` for the code that can read the QChem Hessian from a separate file.

iodata/utils.py
```diff
@@ -37,7 +37,10 @@
 electronvolt: float = 1 / spc.value(u'hartree-electron volt relationship')
 # atomic mass unit (not atomic unit of mass!)
 amu: float = 1e-3 / (spc.value(u'electron mass') * spc.value(u'Avogadro constant'))
-
+# Avogadro constant
+avogadro = spc.value('Avogadro constant')
+# calorie
+calorie = spc.calorie
```
tovrstra: This would be good if IOData would use SI units internally. However, we use atomic units instead. Could you define the following unit, instead of having avogadro and calorie separately?

```python
kcalmol = 1e3 * spc.calorie / spc.value('Avogadro constant') / spc.value('Hartree energy')
```

fwmeng88:
> @fwmeng88 Sorry for the late reply. your message was lost in my inbox. I'll try to clarify a few things here first, because I'm not sure yet how to improve the developer documentation to clarify things.
> 
> A few points that might be helpful:
> 
> * The `mo` attribute contains all the information on the orbitals, that is orbital energies, orbital coefficients, occupation numbers and irrep labels. If some of these are not present, you can leave the corresponding fields to `None`. To make an `mo` attribute, you can create an instance of `MolecularOrbitals`, which is defined here: https://github.com/theochem/iodata/blob/master/iodata/orbitals.py
> * In the output file you included for testing, the basis is not defined in detail, only `BASIS sto-3g`, which can be stored as `obasis_name`, as you already do. You do not have to make an `obasis` object, because the corresponding information is not present in the QChem output.
> 
> Does this help? If you have any further questions, please let me know. I'll make a few comments on the code in an instant.

Thank you for your reply. This is really helpful and I will work on it following the suggestions.

Issue #43, opened 2019-01-28
Infrastructure for making input files for QC codes


tovrstra:
We should have functions to easily write input files for a few quantum chemistry codes of interest (Gaussian, PSI4, ORCA). Suggested API:

```python
from string import Template

# This is just a bare example template, probably lots of features can be added.
default_orca_template = """\
! ${lot} ${basis_name} ${runtype}

*xyz ${charge} ${spinmult}
${atom_lines}
*
"""

def write_input_orca(filename, iodata, template=None):
    """TODO docstring."""
    # some code here to translate iodata attributes to specific names and commands
    # of the QC code. At least the following should be present
    fields = vars(iodata)
    # which can be followed by some code that edits the fields dictionary.
    # For example, you may want to add an 'atom_lines' field that contains a mult-line
    # string with atomic elements, atomic coordinates, and possibly other things like
    # ghost-atom modifiers, etc.
    # Just make sure the values of the fields dictionary are not modified in-place.
    # In some cases, you want to use part of the filename as a field, e.g. to construct
    # the filename of the chk, molden or wfn file.
    # Finally, call a generic routine.
    if template is None:
        template = default_orca_template
    _write_input_generic(filename, fields, template)

def _write_input_generic(filename, fields, template):
    """Write a QC input file.
    
    Parameters
    ----------
    filename
        Filename of the input file to be written.
    fields
        Dictionary of parameters that could potentially be written into the input file.
    template
        A template string.
    """
    with open(filename, 'w') as f:
        f.write(Template(template).substitute(fields))
```

Required attributes for the IOData class are defined in #41.

Some remotely related code can be found here: https://github.com/theochem/horton/blob/master/horton/scripts/atomdb.py

dgasmith:
It would be good to join forces on this, we will begin implementing this kind of tech into two of our codes [Elemental](https://github.com/MolSSI/QCElemental) and [Engine](https://github.com/MolSSI/QCEngine). Both of this will produce and consume schema.

So far we haven't focused too much on the increasing the number of codes that we interface too, but as this ecosystem enters beta that will become a focus. Feel free to join our Slack from any link on the READMEs.

tovrstra:
@dgasmith That could be interesting and I have considered it, but we seem to have requirements that are not covered in QCEngine. Anyway, your comments would be very much appreciated.

[...]

Issue #37, opened 2019-01-25
Clone all io code from molmod


tovrstra:
See:
- https://github.com/molmod/molmod/tree/master/molmod/io
- https://github.com/molmod/molmod/tree/master/molmod/io/test
- https://github.com/molmod/molmod/tree/master/molmod/data/test

evohringer:
Once the draft version of #43 for orca is ready we can collaborate on transferring the io code from molmod and tamkin.

fwmeng88:
Dear Esteban, thank you for work on #43 . Shall we split the tasks of IO functions stolen from `tamkin` and `molmod` to avoid duplicate efforts, please? Except what has been implemented in `iodata`, `cpmd.py` and `gamess.py` both appeared in IO module in `tamkin` and `molmod`.

fwmeng88:
Dear Esteban, I am trying to fix the dependency problem for codes taken from `tamkin`, which is mostly about stealing some codes from `molmod`, https://github.com/theochem/iodata/issues/36#issuecomment-472716912. If we can divide the tasks by clarifying which file formats we are going to take and who is taking care of them, we can get to know the dependet-required codes from `tamkin` and `molmod`. Then we can fix the dependency issuse before taking codes from them. 

The other option is we can add dependency coeds gradually in risk of duplicate work and messy files. Do you have any suggestions or a better way? Thank you. @evohringer 

I can fix those ones soon,
```
from molmod.periodic import periodic
from molmod.io import PunchFile
from molmod import angstrom, amu, calorie, avogadro
from molmod.io.common import SlicedReader
```

tovrstra:
@fwmeng88 Thanks for tackling this. Since you are keeping us updated through this issue, the risk for overlap is small.

We should avoid that molmod becomes a dependency, even temporarily, because it contains outdated code that is not fully inconsistent with similar code in IOData, e.g. the way the units are defined. When you need more units, you can define the here, based on constants from SciPy:

https://github.com/theochem/iodata/blob/master/iodata/utils.py#L38

Similarly, there is some overlap with the following part of iodata too:

https://github.com/theochem/iodata/blob/master/iodata/periodic.py

It would be better to directly use these modules instead of their MolMod counterparts.

[...]

PR #107, opened 2019-05-19
Basic sdf format input/output 

evohringer:
Addresses partially https://github.com/theochem/iodata/issues/43 and https://github.com/theochem/iodata/issues/37


iodata/formats/sdf.py
```diff
@@ -0,0 +1,109 @@
...
+        atcoords[i, 1] = float(words[1]) * angstrom
+        atcoords[i, 2] = float(words[2]) * angstrom
+        try:
+            atnums[i] = sym2num[words[3].title()]
+        except KeyError:
+            atnums[i] = int(words[0])
+    while True:
+        try:
+            words = next(lit)
+        except ValueError:
```
tovrstra:
```suggestion
        except StopIteration:
```

iodata/formats/sdf.py
```diff
@@ -0,0 +1,109 @@
...
+    words = next(lit).split()
+    size = int(words[0])
+    atcoords = np.empty((size, 3), float)
+    atnums = np.empty(size, int)
+    for i in range(size):
+        words = next(lit).split()
+        atcoords[i, 0] = float(words[0]) * angstrom
+        atcoords[i, 1] = float(words[1]) * angstrom
+        atcoords[i, 2] = float(words[2]) * angstrom
+        try:
```
tovrstra: I assume this try-except block is based on the XYZ code, but the XYZ code is actually a poor example of how to use dictionaries in Python. Both should be fixed ideally, to avoid making this suggestion. In general in Python, if there is a way to get the same effect without try-except, then it is preferable for clarity (and speed). In this case, it could also be done like this:

```python
atnum = sym2num.get(words[3].title())
if atnum is None:
    atnum = int(words[3])
atnums[i] = atnum
```

Note that the fallback option should also have index 3. I'm not sure if SDF ever uses integers instead of symbols at index 3, so the fallback option may be redundant anyway? If the fallback is not needed, it could become one line instead:

```python
atnum = sym2num.get(words[3].title(), 0)
```

iodata/formats/sdf.py
```diff
@@ -0,0 +1,109 @@
...
+        'title': title,
+        'atcoords': atcoords,
+        'atnums': atnums
+    }
+
+
+@document_load_many("SDF", ['atcoords', 'atnums', 'title'])
+def load_many(lit: LineIterator) -> Iterator[dict]:
+    """Do not edit this docstring. It will be overwritten."""
+    # SDF files with more molecueles are a simple concatenation of individual SDF files,'
```
tovrstra: typo `molecueles`

iodata/test/test_sdf.py
```diff
@@ -0,0 +1,92 @@
...
+def test_load_dump_consistency(tmpdir):
+    with path('iodata.test.data', 'example.sdf') as fn_sdf:
+        check_load_dump_consistency(tmpdir, fn_sdf)
+
+
+def test_load_many():
+    with path('iodata.test.data', 'example.sdf') as fn_sdf:
+        mols = list(load_many(str(fn_sdf)))
+    assert len(mols) == 2
+    assert mols[1].title == '24978481'
```
tovrstra: Could you add before this line the following?

```
check_example(mols[0])
```

iodata/formats/sdf.py
evohringer: Your are right, only elements are used. I changed it to your suggestion of only line.

iodata/formats/sdf.py
evohringer: Is this because next(lit) is an iterator? I changed it in the commit.

iodata/test/test_sdf.py
evohringer: Done, good idea. Thanks!

evohringer:
I included all changes and left one question about the change of "ValueError" to "StopIteration"

iodata/formats/sdf.py
evohringer: Thanks, corrected

tovrstra:
I just realized we have already a utility to generated truncated files. See https://github.com/theochem/iodata/blob/master/iodata/test/test_cp2k.py#L208 Could you use this? Thanks.

evohringer:
This is a nice feature. Thanks to pointing it out. Its now implemented in the next commit.

tovrstra:
Thank you! Merging...

Issue #78, opened 2019-04-23
Add MOL2 dump function for interaction with 3dmol


tovrstra:
See https://3dmol.csb.pitt.edu/ It should be possible to add cube data somehow, which could be useful for visualizing surfaces in notebooks.

tovrstra:
Fixed by #117

PR #117, opened 2019-05-29
Mol2 format implementation

evohringer:
Implementation of mol2 format. 

Some things I don't like so far is that the first key of the atcharge dictionary is always written out, but I do not know how to tell dump_one function which key (charge type) to write out without changing the api.

Edit by @tovrstra: This adresses parts of #37 and #78.


tovrstra:
@evohringer I agree. One way to solve this would be to pass options into the `load` and `dump` function, for which we have to change the API. This needs some more planning. There are other cases where options to `load` and `dump` would be convenient, e.g. to load and dump extra columns from/to an XYZ file.

Just for now, I would suggest to use the field `'mol2charges'` in the dictionary `'atcharges'`. I'll make suggestions along this lines on the diff.

iodata/formats/mol2.py
```diff
@@ -0,0 +1,152 @@
...
+    print("# Mol2 file created with Iodata", file=f)
+    print("\n\n\n\n\n", file=f)
+    print("@<TRIPOS>MOLECULE", file=f)
+    print(data.title or 'Created with IOData', file=f)
+    print("@<TRIPOS>ATOM", file=f)
+    for i in range(data.natom):
+        n = num2sym[data.atnums[i]]
+        x, y, z = data.atcoords[i] / angstrom
+        out1 = f'{i:6d} {n:2s} {x:10.4f} {y:10.4f} {z:10.4f} '
+        if not data.atcharges:
```
tovrstra:
```suggestion
        atcharges = data.atcharges.get('mol2')
        if atcharges is not None:
```

iodata/formats/mol2.py
```diff
@@ -0,0 +1,152 @@
...
+            # Will write the first atomic charge key of the dictionary
+            charge = data.atcharges[0][i]
```
tovrstra:
```suggestion
            charge = data.atcharges[i]
```

iodata/formats/mol2.py
```diff
@@ -0,0 +1,152 @@
...
+    # MOL2 files with more molecules are a simple concatenation of individual MOL2 files,'
+    # making it trivial to load many frames.
+    while True:
+        try:
+            yield load_one(lit)
+        except IOError:
+            return
+
+
+@document_dump_one("MOL2", ['atcoords', 'atnums'], ['atcharges', 'atffparams',  'title'])
```
tovrstra: Pycodestyle complains because of the double space...
```suggestion
@document_dump_one("MOL2", ['atcoords', 'atnums'], ['atcharges', 'atffparams', 'title'])
```

iodata/formats/mol2.py
evohringer: Changed  and also adapted the part with the atffparams to have the same style.

iodata/formats/mol2.py
evohringer: corrected

iodata/formats/mol2.py
evohringer: changed

evohringer:
Things to be done for being compatible with amber is read in / write out bonds information. Hope to finish it soon and think we should wait to merge once it is ready.

evohringer:
I think now it´s done. 
- Bonds are taken care of 
- Compatibility with antechamber in ambertools19 was checked.

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+            out2 = f'{attype:6s} {1:4d} XXX {charge:14.4f}'
+        else:
+            charge = 0.0000
+            out2 = f'{n:6s} {1:4d} XXX {charge:14.4f}'
+        print(out1 + out2, file=f)
+    if data.bonds is not None:
+        print("@<TRIPOS>BOND", file=f)
+        bonds = data.bonds
+        for i, bond in enumerate(bonds):
+            if bond[2] == 4:
```
tovrstra: Code repetition can be avoided, more or less along the following lines:

```python
bondtype = 'am' if bond[2] == 4 else bond[2]
print(f'{i+1:6d} {bond[0]+1:4d} {bond[1]+1:4d} {bondtype:2s}', file=f)
```

tovrstra:
@evohringer The master branch has now a new feature to easily generate warnings when loading files. 
Instead of calling `print` or `warnings.warn`, one can now do `lit.warn('some warning')`, which will automatically add filename and line number to the warning. I'm updating my comments accordingly. If you want to us `lit.warn` in this PR, you can merge your feature branch with the master branch (provided there are no uncommitted changes) as follow:

```bash
# I'm assuming you're in the mol2_inout branch.
# 1) Fetch the changes from the remote master branch.
# This does not update your local master branch. Only your local copy of origin/master gets updated.
git fetch origin master
# 2) Merge with origin/master
git merge origin/master
# Now you can make additional commits and/or push the branch again to the QCMM repo.
```

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+    """Load element numbers, coordinates and atomic charges."""
+    atnums = np.empty(natoms)
+    atcoords = np.empty((natoms, 3))
+    atchgs = np.empty(natoms)
+    attypes = []
+    for i in range(natoms):
+        words = next(lit).split()
+        # Assume that the first character of atom type is the element
+        try:
+            atnums[i] = sym2num.get(words[5][0].title())
```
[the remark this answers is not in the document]
evohringer: Good point, thanks. I decided to use the atom name in the second column, which I think is more compatible. Normally it is element + number. 

I used your second suggestion with the two characters.

Also added the test with the a new file "silioh3.mol2"

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+                atnums, atcoords, atchgs, attypes = _load_helper_atoms(lit, natoms)
+                atcharges = {"mol2charges": atchgs}
+                atffparams = {"attypes": attypes}
+                result = {
+                    'atcoords': atcoords,
+                    'atnums': atnums,
+                    'atcharges': atcharges,
+                    'atffparams': atffparams
+                }
+                if title is not None:
```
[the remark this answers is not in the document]
evohringer: I agree. Deleted and put into the result dictionary.

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+        print(f'{data.natom:5d} {bonds:6d} {0:6d} {0:6d}', file=f)
+    else:
+        print(f'{data.natom:5d} {0:6d} {0:6d} {0:6d}', file=f)
+    print("@<TRIPOS>ATOM", file=f)
+    for i in range(data.natom):
+        n = num2sym[data.atnums[i]]
+        x, y, z = data.atcoords[i] / angstrom
+        out1 = f'{i+1:7d} {n:2s} {x:15.4f} {y:9.4f} {z:9.4f} '
+        atcharges = data.atcharges.get('mol2charges')
+        attypes = data.atffparams.get('attypes')
```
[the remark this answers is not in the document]
evohringer: Done. Thanks

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+        if atcharges is not None and attypes is not None:
```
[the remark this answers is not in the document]
evohringer: Thanks for the advice. Change accordingly. Still learning :-) The first version was even worse.

iodata/formats/mol2.py
evohringer: Again, thanks. Changed accordingly.

iodata/formats/mol2.py
```diff
@@ -0,0 +1,188 @@
...
+            attype = attypes[i]
+            out2 = f'{attype:6s} {1:4d} XXX {charge:14.4f}'
+        else:
+            charge = 0.0000
+            out2 = f'{n:6s} {1:4d} XXX {charge:14.4f}'
+        print(out1 + out2, file=f)
+    if data.bonds is not None:
+        print("@<TRIPOS>BOND", file=f)
+        bonds = data.bonds
+        for i, bond in enumerate(bonds):
```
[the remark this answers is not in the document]
evohringer: Done

tovrstra:
Thanks!

PR #119, opened 2019-05-31
Remove some print lines and use warnings module where needed.

tovrstra:
Warnings are now also tested in the unit tests. This is related to the discussion in #117.


tovrstra:
P.S. By checking warnings, I discovered we were not testing the right file in test_molden.py, line 225. This is neat.

tovrstra:
Thanks, going in.

PR #120, opened 2019-06-01
Add method for producing basis object compatible with gbasis

kimt33:
This PR adds compatibility to https://github.com/theochem/gbasis.

1. Add `gbasis_basis` to `IOData`
2. Add test

The method is added to IOData instead of MoleculeBasis or Shell because the
coordinates are not stored within either of these classes. Instead, the index is
stored, and we need to refer back to the IOData to use instances of these
classes. This seems to suggest that these classes will only be used within the 
IOData class and that they serve only to store information. Of course, we can move 
this method to the MoleculeBasis class and pass the coordinates, but that seems 
messy (and unsafe even). 

Though gbasis supports spherical contractions, the implemented wrapper does not
because the current documentation ordering of spherical primitives (in
`MolecularBasis`) does not have sufficient information to do so (for me
anyways). Spherical contractions can be supported later.

### DIFF
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -22,7 +22,7 @@
 import attr
 import numpy as np
 
-from .basis import MolecularBasis
+from .basis import MolecularBasis, Shell
 from .orbitals import MolecularOrbitals
 from .utils import Cube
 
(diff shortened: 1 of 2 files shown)


kimt33:
[Add wrapper to iodata](https://trello.com/c/Z5mAU4Ek/53-add-wrapper-to-iodata)

kimt33:
Hmm. I'm not sure what to do with the gbasis dependency. I think it should be added to setup.py as an optional dependency and travis.yml tweaked to install the optional dependency. Either way, gbasis should be up in pypi for this to be easily done.

tovrstra:
@kimt33 We intended to have no dependencies of IOData on other packages. I'll look into this soon but might not be able to do so todya.

kimt33:
@tovrstra do you mean that IOData will not depend on any other packages or that the other packages will not depend on IOData? I'm guessing the former. I just figured that since IOData acts as the link between different packages, that it will also interface to other software packages (just as it interfaces to gaussian log files and whatnot). If not, then, I'm guessing, each software will need to explicitly support IOData in order to use IOData?

Or did you mean specifically that IOData will have minimal dependencies? So it can return arguments that can easily be passed into classes/functions in other packages?

kimt33:
@tovrstra but it's no problem this isn't the right location for this code. I can move this code to gbasis if that's better.

kimt33:
Appropriate parts were moved to the gbasis module. See https://github.com/theochem/gbasis/pull/69

PR #123, opened 2019-07-02
Modification of cube files writing function

thomaspigeon:
Adding some missing hard returns in the function _write_cube_data

### DIFF
--- a/iodata/test/test_cube.py
+++ b/iodata/test/test_cube.py
@@ -66,6 +66,26 @@ def test_load_dump_load_aelta(tmpdir):
     dump_one(mol1, fn_cube2)
     mol2 = load_one(fn_cube2)
 
+    with open(fn_cube2) as f:
+        line_counter = 0
+        block_counter = 0
+        for line in f:
+            line_counter += 1
+            if line_counter > 6 + len(mol2.atnums):
+                if mol2.cube.shape[2] % 6 == 0:
+                    assert len(line.split()) == 6
+                if mol2.cube.shape[2] % 6 != 0:
+                    block_line_counter = line_counter - (
+                        6
+                        + len(mol2.atnums)
+                        + block_counter * (mol2.cube.shape[2] // 6 + 1)
+                    )
+                    if 1 <= block_line_counter <= mol2.cube.shape[2] // 6:
+                        assert len(line.split()) == 6
+                    if block_line_counter == mol2.cube.shape[2] // 6 + 1:
+                        assert len(line.split()) == mol2.cube.shape[2] % 6
+
+
     assert mol1.title == mol2.title
     assert_allclose(mol1.atcoords, mol2.atcoords, atol=1.e-4)
     assert_equal(mol1.atnums, mol2.atnums)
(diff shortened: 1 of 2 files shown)


tovrstra:
@thomaspigeon Could you fix the following:

```
171:48    iodata/formats/cube.py  E225 missing whitespace around operator
```

See https://travis-ci.com/theochem/iodata/jobs/212774176#L641

Just out of interest, which programs can't read cube files without the return characters you added? Is there a way to test this?

thomaspigeon:
I just added a commit to fix it and prevent double hard returns.
I had issues with visualizing the cube file with Gabedit.
I don't know how to test this, when I read then rewrote a cube file under another name and read this rewritten file with IOData I ended up having two identical object of IOData class. By the way the test file iodata/test/data/aelta.cube would not be suitable for this kind of test as the "block_size" is a multiple of 6 (=12) so that makes no differences.

tovrstra:
Thanks for all the extra info, all makes sense now. Could you add a check for empty lines in the following test:

https://github.com/theochem/iodata/blob/master/iodata/test/test_cube.py#L61

Something like this should do the trick:

```python
with open(fn_cube2) as f:
    for line in f:
        assert len(line.strip()) != 0
```

Other than that, no comments. @FarnazH  can you also take a look? I believe you have been experimenting with the cube file code in Chemtools recently?

tovrstra:
I've clicked a few wrong buttons in Github causing a lot of clutter. Sorry for that.

thomaspigeon:
I just added a few assert to check for the format of each line of the volumetric data part.

thomaspigeon:
I am really sorry but I don't understand the errors in Travis log, I tried twice to fix it but still I don't get what's wrong.

iodata/test/test_cube.py
```diff
@@ -66,6 +66,24 @@ def test_load_dump_load_aelta(tmpdir):
...
+        block_counter = 0
+        for line in f:
+            line_counter += 1
+            if line_counter > 6 + len(mol2.atnums):
+                if mol2.cube.shape[2] % 6 == 0:
+                    assert len(line.split()) == 6
+                if mol2.cube.shape[2] % 6 != 0:
+                    block_line_counter = (
+                            line_counter - (6 + len(mol2.atnums)
+                            + block_counter * (mol2.cube.shape[2] // 6 + 1))
```
tovrstra: This is the piece of code to which the error messages apply.

```
79:29     iodata/test/test_cube.py  E126 continuation line over-indented for hanging indent
80:29     iodata/test/test_cube.py  E128 continuation line under-indented for visual indent
```

The code on these two lines is indented incorrectly. There are many possible ways to fix this. If you have cardboardlint installed, just run `cardboardlinter -F`. For details see https://iodata.readthedocs.io/en/latest/contributing.html

This should be a correct indentation, which is also fairly readable.

```python
                    block_line_counter = (
                        line_counter - (
                            6 + len(mol2.atnums) +
                            block_counter * (mol2.cube.shape[2] // 6 + 1)
                        )
                    )
```

The basic rule is to always indent by four spaces when breaking a line right after an opening bracket (`[`, `(`, ...). More indentation should be added whenever wrapping lines after new opening brackets. When wrapping not directly after a bracket, indentation should match character after the last opening bracket that was not closed on the previous line. (I'm not sure if that is even proper English. I hope it makes sense.) These indentation rules are rather complicated at first but they can improve readability a lot.

thomaspigeon: Thank you a lot for your explanations. I think I better understand it now, it should be ok now.

PR #125, opened 2019-07-05
Add acknowledgements section to main page

tovrstra:
Note that the latest development version of the docs can always be found here: https://iodata.readthedocs.io/en/latest/

It will be updated with this acknowledgements section as soon as this PR gets merged.


PaulWAyers:
It looks good to me. I think it's best to be vague about our "ongoing sources of funding" because it would be annoying to have to constantly update grant numbers and things like that.

tovrstra:
All should be ok now.

PR #127, opened 2019-07-08
Add`build_obasis` and `load_one` for WFX 

fwmeng88:
- [x] Add `build_obasis`
- [x] Add `load_one`
- [x] Fix pylint
- [x] Add tests for `load_one`
- [x] Docstring cleanups


tovrstra:
I just took a quick look and have some suggestions.

- The API of the `load_one` function is not yet compatible with the functions in `iodata.api`. You can look into the other file format modules for examples.

- It may be easier to combine `load_one` and `load_wfw_low` into one function. The large number of return values of `load_wfx_low` is not very practical. The `load_one` function returns a dictionary, which can be easier to handle, also in the tests.

- You may also store extra information in the `extra` attribute of `IOData`. It is a dictionary where you are free to put any additional information loaded from a file, for which there are no documented attributes for IOData.

- When testing `load_one`, can you use the functions in `iodata.api`? This way, the tests make sure the wfx module is compatible with the general interface in `iodata.api`.

Thanks!

fwmeng88:
> In addressing @tovrstra comments, I think it's better to remove `loaf_wfx_low` at the last stage after making sure that everything works fine. Also, can you please make sure to add tests for the new functions?

Thank you for the suggestions. I will try to add tests for the functions, which is also my intention.

fwmeng88:
> I just took a quick look and have some suggestions.
> 
> * The API of the `load_one` function is not yet compatible with the functions in `iodata.api`. You can look into the other file format modules for examples.
> * It may be easier to combine `load_one` and `load_wfw_low` into one function. The large number of return values of `load_wfx_low` is not very practical. The `load_one` function returns a dictionary, which can be easier to handle, also in the tests.
> * You may also store extra information in the `extra` attribute of `IOData`. It is a dictionary where you are free to put any additional information loaded from a file, for which there are no documented attributes for IOData.
> * When testing `load_one`, can you use the functions in `iodata.api`? This way, the tests make sure the wfx module is compatible with the general interface in `iodata.api`.
> 
> Thanks!

Thank you very much for your suggestions. I am OK with all the items except the first one. For the `load_one` in `iodata/api.py`,
https://github.com/theochem/iodata/blob/66c0553b3b18331a5b2279b1888c86607b377c64/iodata/api.py#L102-L107
the function selects the format and the corresponding `load_one` function in each format, then it builds `lit` as the argument of `IOData(**format_module.load_one(lit))`, where I use the file name instead. Can I make a little modification on `load_one` in `iodata/api.py` by adding an if statement? @tovrstra @FarnazH 

```python
def load_one(filename: str, fmt: str = None) -> IOData:
    """Load data from a file.
    This function uses the extension or prefix of the filename to determine the
    file format. When the file format is detected, a specialized load function
    is called for the heavy lifting.
    Parameters
    ----------
    filename
        The file to load data from.
    fmt
        The name of the file format module to use. When not given, it is guessed
        from the filename.
    Returns
    -------
    out
        The instance of IOData with data loaded from the input files.
    """
    fmt_fname = {'wfx'}
    format_module = _select_format_module(filename, 'load_one', fmt)
    if fmt in fmt_fname:
        try:
             return IOData(**format_module.load_one(filename))
        except IOError:
            IOError("Errors were found in inpunt file.")
    else:
        lit = LineIterator(filename)
        try:
            return IOData(**format_module.load_one(lit))
        except StopIteration:
            raise lit.error("File ended before all data was read.")
```

FarnazH:
@fwmeng88 the `load_one` in `iodata/api.py` should not change at all. It is designed to work with all formats. Like other formats, the functions in `formats/wfx.py` need to change to take `LineIterator` as an argument. Please let me know if there is any ambiguity.

fwmeng88:
Thank you for the useful comments! @FarnazH 

I have updated `wfx.py` using `LineIterator` as an argument. But one thing I have been working for a  while, which is about the iterator. All the private functions work perfectly when only input the `LineIterator` alone, but when we put them together, `load_wfx_low` failed. I have checked and I think it is because of the way `_helper_load_section` reads the `lit`. Sometimes, we don't know how many lines one section can have. Therefore, I used `while` loop and the function breaks when finishing reading `lit`. If `_helper_load_section` cannot find the property it wants, it keeps reading and consumes up the iterator. In this PR, the order of reading properties is not following standard documentation, which makes the next operation cannot use the initial iterator. For example, 

```python
lit = LineIterator('water_sto3g_hf.wfx')
# string type properties
title, keywords, model_name = _helper_str(lit).values()
# todo: Check keywords
assert (keywords[0] in ['GTO', 'GIAO', 'CGST']), \
    "The keywords should be one out of GTO, GIAO and CGST."
num_atoms = _helper_int(lit, proper_name='num_atoms')
```
this code block will try to find the information about *title, keywords, mode_name*, but later `lit` has been used up. When `_helper_int` is trying to find the information about the number of atoms, it fails. When we comment `title, keywords, model_name = _helper_str(lit).values()`, we can find the right information about `num_atoms`.

I tried to follow the order of the standard file when extracting the information, but some information is not present. In this case, it will use up the iterator. For example, *Model* is in https://github.com/theochem/iodata/blob/master/iodata/test/data/water_sto3g_hf.wfx, not https://github.com/theochem/iodata/blob/master/iodata/test/data/h2_ub3lyp_ccpvtz.wfx. The following operation will run short of the 'whole' iterator, resulting in errors.

If you run the pytest, you will notice the problem. Can you suggest a good way to deal with this? Thank you very much! @tovrstra @FarnazH

FarnazH:
@fwmeng88 In fixing the issue you raised, I refactored the code for parsing and processing the data which also removed the functions with many returned values as @tovrstra suggested.

There are now just two functions, `parse_wfx` that reads all the sections available in a WFX file (it makes sure that each section has proper starting and ending headline, but it doesn't look for specific sections as unfortunately, information recorded in a WFX can vary). Then `load_data_wfx` puts the parsed data into the proper type and format. I fixed the two tests on water & H2.

iodata/formats/wfx.py
```diff
@@ -18,224 +18,156 @@
...
-            dict_int[key] = np.array(int_info[0], dtype=np.int)
+    # load raw data
+    data = parse_wfx(lit)
+    # process raw data
+    result = {}
+    for key, value in data.items():
+        if key in labels_str:
+            result[labels_str[key]] = value[0]
+        elif key in labels_number:
+            result[labels_number[key]] = np.array(eval(value[0]))
```
tovrstra: I'd suggest (like pylint) to replace `eval` by `int` or `float`. This can be done by splitting `labels_number` into `labels_int` and `labels_float`, with corresponding changes in the `elif` branches.

tovrstra: Also, if these are not arrays but just numbers, the `np.array()` call can be omitted.

iodata/formats/wfx.py
```diff
@@ -18,224 +18,156 @@
...
+        elif key in labels_array:
+            breakdown = [eval(item) for line in value for item in line.split()]
```
tovrstra: Idem: better to avoid calling eval on data from external sources. The solution here is similar, i.e. splitting `labels_array` into `labels_array_int` and `labels_array_float`.

Instead of the list comprehension, also `np.fromstring(line, dtype=..., sep=" ")` can be used. See https://mail.google.com/mail/u/0/#inbox That would be a lot faster for large arrays.

iodata/formats/wfx.py
```diff
@@ -18,224 +18,156 @@
...
+            result[labels_array[key]] = np.array(breakdown)
+        elif key in labels_other:
+            result[labels_other[key]] = value
         else:
-            dict_int[key] = np.array(None)
-
-    return dict_int
+            warnings.warn("Not recognized label, skip {0}".format(key))
 
+    # reshape some arrays
```
tovrstra: Instead of the `_check_tag` function, a simple check can be added before this line. For example, one could verify if all the values defined line the `labels_*` dictionaries are also present in the `results` dictionary. E.g. like

```python
for labels_dict in labels_str, labels_int, labels_float, labels_array_int, labels_array_float, labels_other:
    for name in labels_dict.values():
        if name not in result:
            raise IOError("some message")
```

tovrstra:
Looks good, safe for few minor comments above. It may be worth checking which data from the results dictionary is not used yet, to store it in the `IOData.extra` dictionary or in other attributes.

iodata/formats/wfx.py
```diff
@@ -18,224 +18,156 @@
...
+    result['atcoords'] = result['atcoords'].reshape(-1, 3)
+    result['mo_coeff'] = result['mo_coeff'].reshape(result['num_primitives'], -1, order='F')
+    # process mo spin type
+    mo_spin_list = [i.split() for i in result['mo_spin']]
+    mo_spin_type = np.array(mo_spin_list, dtype=np.unicode_).reshape(-1, 1)
+    result['mo_spin'] = mo_spin_type[mo_spin_type[:, 0] != 'and']
+    # process nuclear gradient
+    gradient_mix = np.array([i.split() for i in result.pop('nuclear_gradient')]).reshape(-1, 4)
+    result['gradient_atoms'] = gradient_mix[:, 0].astype(np.unicode_)
+    result['gradient'] = gradient_mix[:, 1:].astype(float)
```
tovrstra: `gradient_atoms` does not get used and I'm not sure why it is present in the file format? Can it happen that nuclear forces are absent for some atoms? If not, the above three lines can be replaced by

```python
result['gradient'] = np.array([line.split()[1:] for line in result.pop('nuclear_gradient')], dtype=float)
```
(shape should be ok)

iodata/formats/wfx.py
```diff
@@ -284,3 +216,148 @@ def _check_tag(f_content: str):
...
+            data['mo_occ'], data['mo_coeff'], data['mo_energy'], None)
+
+    result = {
+        'title': data['title'],
+        'atcoords': data['atcoords'],
+        'atnums': data['atnums'],
+        'obasis': obasis,
+        'mo': mo,
+        'energy': data['energy'],
+    }
```
tovrstra: It would be nice to add `'atgradient': data['gradient']`. There may be other useful additions.

fwmeng88: Sure, I will do.

fwmeng88:
Thank you very much for the help with WFX! @FarnazH @tovrstra The new code is very clean and efficient. 
I will work on it and push the updated codes to this PR.

iodata/formats/wfx.py
fwmeng88: I have fixed it following your suggestions. Thank you for the advice.

fwmeng88: BTW, is it because `eval` is more used for evaluating the python strings, but `int` and `float` are more specific for converting strings into numbers? For example, `eval` can be used for `eval('8 + 9')`.

iodata/formats/wfx.py
fwmeng88: Fixed.

iodata/formats/wfx.py
tovrstra: `eval` is indeed a much more advanced function. The main reason for avoiding it is the security risks involved. When you apply `eval` to data read from a file, one can easily inject code from a malicious data file to execute anything. For example, this could be a real risk when someone uses IOData to process wavefunctions uploaded to a web server.

fwmeng88: Thank you very much! A new technique learned.

iodata/formats/wfx.py
fwmeng88: I have fixed by adding some lines of codes with a similar pattern.

iodata/formats/wfx.py
fwmeng88: The nuclear gradient in WFX is composed of gradient atoms (`gradient_atoms`) and the corresponding gradient (`gradient`). Because not all nuclear have gradient data, we need to keep them. I tried to make them a dictionary,
```python
    result = {
        'title': data['title'],
        'atcoords': data['atcoords'],
        'atnums': data['atnums'],
        'obasis': obasis,
        'mo': mo,
        'atgradient': {'gradient_atoms': data['gradient_atoms'], 'gradient': data['gradient']},
    }
```
Will this be OK?

iodata/formats/wfx.py
fwmeng88: Fixed.

iodata/formats/wfx.py
tovrstra: That will not work because `atgradient` has to be an `(natom,3)` array. Also adding new undocumented attributes, like `gradient` will not work. We have such restrictions in place to make sure that data loaded from different files can be used in the same way, independently of the file format. I would suggest the following:
- Remove `gradient` from the dictionary
- Turn `atgradient` into a numpy array with shape `(natom, 3)` and fill in as much elements as possible, leaving undefined ones `NaN`.

tovrstra: Sorry, I misread your code. Let my try to clarify. The following would be good:

```python
result = {
        # some lines omitted
        'atgradient': atgradient,
        # some lines omitted
    }
```

where `atgradient` is simply a numpy array with shape `(natom, 3)`, where as much as possible values are filled in, leaving the rest to `NaN`

iodata/formats/wfx.py
```diff
@@ -56,8 +56,10 @@ def load_data_wfx(lit: LineIterator) -> dict:
         '<Number of Alpha Electrons>': 'num_alpha_electron',
         '<Number of Beta Electrons>': 'num_beta_electron',
         '<Number of Primitives>': 'num_primitives',
-        '<Electronic Spin Multiplicity>': 'spin_multi',
```
FarnazH: This is a minor remark, but here it is better to keep the `,` after the last element of a sequence. In this way, when someone adds a new element to the sequence, the last line doesn't need to change for adding the `,`. This makes the changes shown by git more meaningful.

iodata/test/data/water_unrecognized_tag.wfx
```diff
@@ -0,0 +1,156 @@
+<Haha>
+This is a unrecognizable tag
+</Haha>
```
FarnazH: I don't think we need a full wfx file to test an unexpected tag. It's better to add a smaller test file.

iodata/formats/wfx.py
```diff
@@ -332,12 +332,26 @@ def load_one(lit: LineIterator) -> dict:
             'unrestricted', norba, norb - norba,
             data['mo_occ'], data['mo_coeff'], data['mo_energy'], None)
 
+    data_extra = {}
+    extra_list = ['keywords', 'model_name', 'num_perturbations', 'num_electrons',
```
FarnazH: Not all the items listed in `extra_list` should be added `extra` item returned. Some of them like `nume_primitives` is already stored in `obasis` or `num_electrons`, `num_alpha_electron` and `num_beta_electron` that are already stored in `mo`.

iodata/formats/wfx.py
```diff
@@ -109,6 +109,36 @@ def load_data_wfx(lit: LineIterator) -> dict:
         else:
             warnings.warn("Not recognized label, skip {0}".format(key))
 
+    # check required properties in WFX
+    required_prop = {'<Title>': 'title',
```
FarnazH: The items listed in `required_prop` dictionary already exists in `labels_*` dictionaries, so I think this is a redundant dictionary. Also, only checking section tags is enough, so there is no need to use `prop_name` as well.

iodata/test/data/h2o_error.wfx
```diff
@@ -56,7 +56,7 @@ Restricted HF
 1 1 1 1 1 2 2 2 3 3
 3
 </Primitive Centers>
-####<Primitive Types>
+<Primitive Types>
```
FarnazH: Can you please clarify why this line was changed? I believe it was intended to test the error raised when `</Primitive Types>` end tag exists, but the starting tag does not exist...

iodata/formats/wfx.py
fwmeng88: Thank you for the clarification and Farnaz's codes works.

iodata/formats/wfx.py
fwmeng88: I never thought about this, but this is definitely necessary. I have followed this pattern for the coming commits. Thank you for pointing it out and I learned something new.

iodata/test/data/h2o_error.wfx
fwmeng88: At that stage, the WFX module can only check if the head and tail match. For example, `FileFormatError: test.wfx:61 Expected line </Primitive Centers> but got </Primitive Centers 123>` when the tail is "</Primitive Centers 123>". But when we are using "####</Primitive Centers>", it will raise `OSError: The <Primitive Centers> section is missing!`. However, this section is there and just the wrong tail. So I think this is not a typical error and just make it correct.

iodata/formats/wfx.py
```diff
@@ -295,7 +295,7 @@ def load_one(lit: LineIterator) -> dict:
     data = load_data_wfx(lit)
 
     # Build the basis set and the permutation needed to regroup shells.
-    obasis, permutation = build_obasis(data['icenters'], data['type'], data['exponents'])
+    obasis, permutation = build_obasis(data['centers'], data['types'], data['exponents'])
```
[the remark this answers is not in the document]
fwmeng88: Now I understand why there is an indexerror. Thanks.

iodata/test/test_wfx.py
```diff
@@ -280,6 +280,10 @@ def test_load_wfx_low_wrong_tail_h2o():
             load_data_wfx(lit)
     assert error_info.match(r"Expected line .* but got .*")
 
+def test_load_one_h2o():
+    """Test load_one with h2o sto-3g WFX input."""
+    data = load_one('iodata/test/data/water_sto3g_hf.wfx')
+    assert data['title'] == 'H2O HF/STO-3G//HF/STO-3G'
```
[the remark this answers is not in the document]
fwmeng88: No problem. Get it fixed.

tovrstra:
Some tests still fail, related to orthonormality of mo's. I'll look into this later.

tovrstra:
It should be working now, but a few more (small) test files should be included, e.g. an unrestricted wavefunctions with different alpha and beta orbitals. We should also check for other variations of WFX files that can be made with AIMALL.

fwmeng88:
Dear Toon, thank you so much!  The codes seem to be very good shape. Thank you for fixing the test codes again.

tovrstra:
Thanks everyone. Additional improvements, if any, are always welcome in future PRs. I'll merge this one now because it seems to work and this PR has been open for so long.

Issue #101, opened 2019-05-13
Disable all type-checking features of pylint and use mypy instead


tovrstra:
See #100 for explanation.

PR #128, opened 2019-07-20
Support custom columns in XYZ files

### LINKED ISSUE #118: Make it possible to pass options to load and dump functions
Adding arbitrary optional arguments tp `load` and `dump` would make it possible to deal with variations of file formats, e.g. extra columns in XYZ files. Another use case can be found in #117 . 

We could add `**kwargs` to the `load_*` and `dump_*` functions and pass them on to corresponding functions in the file format modules. 

@FarnazH Do you know of other relevant use cases?

tovrstra:
Fixes #118.

Some other notable points:
- The PR includes an example of this feature for XYZ files.
- Exceptions were added to `.pylintrc` due to false-positives, which seem fairly general. Mipy should be used to catch such problems, as explained in #101.
- The documentation page generator `gen_formats.py` was extended (and simplified) to list the optional arguments supported by each file format.
- This is not API breaking, but rather an extension.


iodata/formats/xyz.py
```diff
@@ -41,55 +71,99 @@
...
+                dest = np.zeros((natom,) + shapesuffix, dtype=dtype)
+                if keyname is None:
+                    data[attrname] = dest
+                else:
+                    data.setdefault(attrname, {})[keyname] = dest
+                dest = dest[:1]
+            else:
+                # For all remaining atoms, the array with atomic properties is
+                # just retrieved.
+                if keyname is None:
```
evohringer: I´m not an expert in python, but it is hard to understand how this works. Could you explain it a little bit better?

tovrstra: I've made a few changes and added more comments. Can you check if there is anything specific that is not obvious?

tovrstra:
Thanks. I'll merge it.

PR #129, opened 2019-07-26
Implementation of pdb reading and writing features.

evohringer:
Hi, that is a basic implementation which could be extended with the new optional reading and writing features from @tovrstra


iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+                nums.append(sym2num.get(symbol, sym2num.get(symbol[0], None)))
+            except ValueError:
+                # When the position 77:79 does not exist also use the atom type
+                symbol = line[13:16].title()
+                nums.append(sym2num.get(symbol, sym2num.get(symbol[0], None)))
+            x = float(line[30:38])
+            y = float(line[38:46])
+            z = float(line[46:54])
+            coords.append([x, y, z])
+            attypes.append(line[13:16].strip())
```
tovrstra: The range `13:16` from the PDB spec is using different conventions than Python slices. In the PDB spec, the last index is inclusive and they start counting from 1 instead of zero. (See http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM) This is terribly confusing and I believe the following would be correct:

```suggestion
            attypes.append(line[12:16].strip())
```
(Please check carefully.)

Similar changes may be needed elsewhere in this pull request.

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+            restypes.append(line[18:20].strip())
+            molecule_found = True
+        if line[0:3] == "END" and molecule_found:
+            atnums = np.array(nums)
+            atcoords = np.array(coords) * angstrom
+            attypes = tuple(attypes)
```
tovrstra: It would be interesting to use an array of strings here, to enable numpy operations. This can be done as follows:
```suggestion
            attypes = np.array(attypes, dtype="U4")
```
With this dtype, every array element is a unicode string with at most four characters.

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+            restypes = tuple(restypes)
```
tovrstra: Similar suggestion:
```suggestion
            restypes = np.array(restypes, dtype="U3")
```

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+            break
+    if molecule_found is False:
+        raise lit.error("Molecule could not be read")
+    return result
+
+
+@document_load_many("PDB", ['atcoords', 'atnums', 'atffparams'], ['title'])
+def load_many(lit: LineIterator) -> Iterator[dict]:
+    """Do not edit this docstring. It will be overwritten."""
+    # PDB files with more molecules are a simple concatenation of individual MOL2 files,'
```
tovrstra: Typo
```suggestion
    # PDB files with more molecules are a simple concatenation of individual PDB files,'
```

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+            yield load_one(lit)
+        except IOError:
+            return
+
+
+@document_dump_one("PDB", ['atcoords', 'atnums'], ['atffparams', 'title'])
+def dump_one(f: TextIO, data: IOData):
+    """Do not edit this docstring. It will be overwritten."""
+    print(str("TITLE    " + data.title) or "TITLE    Created with IOdata", file=f)
+    attypes = data.atffparams.get('attypes')
```
tovrstra: This would generate a `ValueError` when `attypes` is not present. Suggested fix:
```suggestion
    attypes = data.atffparams.get('attypes', None)
```
To test this case, you could add a unit test that loads an XYZ file and writes it back out as PDB file to a temporary directory. You could even load the PDB again to check if it has the expected contents.
(Same comment for next line.)

iodata/formats/pdb.py
evohringer: Thanks, I corrected all the indices at well

iodata/formats/pdb.py
evohringer: The reason I used a tuple is that I thought it expects it to be a tuple when you assign it to the atffparams dictionary.

iodata/formats/pdb.py
evohringer: same as below

iodata/formats/pdb.py
evohringer: I added a test for the attype and restype read and write from an xyz file.

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+        if line[0:4] == "ATOM" or line[0:6] == "HETATM":
+            try:
+                # Try reading element from position 77:79 referenced in pdb format
+                words = line[77:79].split()
+                if not words:
+                    # If not guess it from the atom type
+                    words = line[13:16].split()
+                symbol = words[0].title()
+                nums.append(sym2num.get(symbol, sym2num.get(symbol[0], None)))
+            except ValueError:
```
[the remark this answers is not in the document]
evohringer: Done. Thanks!

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+    restypes = []
+    molecule_found = False
+    while True:
+        try:
+            line = next(lit)
+        except StopIteration:
+            break
+        if line.startswith("TITLE") or line.startswith("COMPND"):
+            words = line.split()
+            title = words[1:]
```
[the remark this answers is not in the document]
evohringer: I added .rstrip() to get read of the newline character. I will leave it to read only one line.

iodata/formats/pdb.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            title = "Pdb file from Iodata"
```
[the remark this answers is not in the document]
evohringer: Done

evohringer:
> Thanks! I've made lots of small requests. Most should be fixable with little work. I would not bother supporting multi-line titles at this stage yet. We can always add more features in later pull requests. Another interesting extension would be to load the connectivity, for which we have a corresponding field in IOData, see
> 
> https://github.com/theochem/iodata/blob/58d5e4bd1fb83409ca1e6ceeb474c4358c6ddc10/iodata/iodata.py#L105-L109
> 
> 
> Anway, this is also something for another pull request.

- [ ] I will add bonds in a feature pull request as you suggested. I had this in mind.

tovrstra:
I've restarted the failed tests because the error seemed unrelated to this pull request.

tovrstra:
@evohringer The following is not a big deal but I guess you would like to know. Your email address is not set in your git configuration. This can be fixed as follows:

```bash
git config --global user.email '<email here>'
```

Your current email address in the commits is evohringer@Estebans-MacBook-Pro.local, which is probably not intended. See https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup for details.

iodata/formats/pdb.py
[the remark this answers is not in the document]
tovrstra: sounds good

iodata/formats/pdb.py
tovrstra: great!

iodata/formats/pdb.py
tovrstra: The `IOData` class will not enforce a specific type for the values of the `atffparams` dictionary. Any type can be used in principle, but the documentation says the values of this dictionary are arrays: https://github.com/theochem/iodata/blob/58d5e4bd1fb83409ca1e6ceeb474c4358c6ddc10/iodata/iodata.py#L80-L85

I can't think of a use case where tuples would have an advantage over arrays, but if there is a technical reason, please let me know. Arrays would in practice allow for faster operations, e.g. displacing all atoms of a given type would be relatively fast and easy with arrays: 

```python
mol.atcoords[mol.atffparams["attypes"] == "CA"] += np.array([1.0, 0.0, 0.0])
```

iodata/formats/pdb.py
tovrstra: looks good!

iodata/formats/pdb.py
evohringer: You are right. I had previously tried with a list if I remember correctly and that gave an error. I changed both restypes and attypes to arrays.
Thanks

iodata/formats/pdb.py
evohringer: Changes to arrays

Issue #131, opened 2019-07-29
Document basis functions with equations


tovrstra:
The following should be covered:

- pure and Cartesian functions
- sign conventions
- normalization conventions

tovrstra:
This was originally brought up in #122.

tovrstra:
See https://en.wikipedia.org/wiki/Solid_harmonics#Real_form

tovrstra:
This should also cover the string conventions used to define the order and sign of basis functions.

PR #133, opened 2019-07-31
Simplify strings in conventions for pure functions

tovrstra:
The character for the angular momentum is omitted because it is redundant. With this change, the conventions are no longer limited in angular momentum by the availability of alphabetical characters.

Some mistakes in the docstring of `MolecularBasis` are also fixed. More elaborate documentation will be added in a separate PR. See #131

### DIFF
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -150,16 +150,28 @@ class MolecularBasis(NamedTuple):
         .. code-block:: python
 
             {
+                ### Conventions for Cartesian functions
+                # E.g., alphabetically ordered Cartesian functions.
                 (0, 'c'): ['1'],
                 (1, 'c'): ['x', 'y', 'z'],
-                # alphabetically ordered Cartesian functions
                 (2, 'c'): ['xx', 'xy', 'xz', 'yy', 'yz', 'zz'],
-                # or Wikipedia-ordered real solid spherical harmonics
-                # c = cosine-like
-                # s = sine-like
-                (2, 'p'): ['dc2', 'dc1', 'dc0', '-ds1', '-ds2'],
-                ...
+                ### Conventions for pure functions.
+                # The notation is referring to real solid spherical harmonics.
+                # See https://en.wikipedia.org/wiki/Solid_harmonics#Real_form
+                # c = cosine of the azimuthal angle
+                # s = sine of the azimuthal angle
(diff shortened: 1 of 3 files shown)


iodata/basis.py
```diff
@@ -150,17 +150,29 @@ class MolecularBasis(NamedTuple):
...
-                # or Wikipedia-ordered real solid spherical harmonics
-                # c = cosine-like
-                # s = sine-like
-                (2, 'p'): ['dc2', 'dc1', 'dc0', '-ds1', '-ds2'],
-                ...
+                ### Conventions for pure functions.
+                # The notation is referring to real solid spherical harmonics.
+                # See https://en.wikipedia.org/wiki/Solid_harmonics#Real_form
+                # c = cosine of the azimuthal angle
+                # s = sine of the azimuthal angle
```
kimt33: Would it be alright to add `number = magnetic quantum number` (or the correct term) here? It would be helpful to have a definition of each component of the notation.

iodata/basis.py
```diff
@@ -150,17 +150,29 @@ class MolecularBasis(NamedTuple):
...
+                # For example, wikipedia-ordered real spherical harmonics,
+                # see https://en.wikipedia.org/wiki/Spherical_harmonics#Real_form
+                (2, 'p'): ['s2', 's1', 'c0', 'c1', 'c2'],
+                # Different quantum-chemistry codes may use incompatible
+                # orderings and sign conventions. E.g. Molden files written
+                # by Orca use the following convention for pure f functions:
+                (3, 'p'): ['c0', 'c1', 's1', 'c2', 's2', '-c3', '-s3'],
+                # Note the sign change of the last two basis functions.
```
kimt33: Could this elaborate more on the sign change? Something like `Note that the minus sign in the last two basis functions denotes that the signs of these harmonics have been changed.`

I thought it'd be nice to have a more explicit definition, but maybe it's a bit wordy and is redundant. So whatever you think is best.

iodata/basis.py
tovrstra: good point.

iodata/basis.py
tovrstra: I'll add it. I'm going to write out all the equations and details in documentation in the next PR, but it does not hurt to make it a bit more explicit here.

PR #138, opened 2020-01-29
(WIP) Experiment with type checking and fix bug

tovrstra:
This fixes a small bug found when working on theochem/gbasis#78.

The main purpose of this PR is discuss how we can introduce more type checking into IOData. If we had type checking as implemented in this PR, it would have prevented the fixed bug, so it seems the right thing to do. We can apply the same technique to other parts of IOData, but I first want to make sure this will not cause problematic side effects. Such changes would more or less involve the following:

- Replace NamedTuple classes by regular classes decorated with [attr](https://pypi.org/project/attrs/) machinery.
- Add validators to some fields to impose types or other requirements where it makes sense.
- Possibly fix bugs detected by the validators.

### DIFF
--- a/iodata/attrutils.py
+++ b/iodata/attrutils.py
@@ -1,0 +1,107 @@
+# IODATA is an input and output module for quantum chemistry.
+# Copyright (C) 2011-2019 The IODATA Development Team
+#
+# This file is part of IODATA.
+#
+# IODATA is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 3
+# of the License, or (at your option) any later version.
+#
+# IODATA is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>
+# --
+"""Utilities for building attr classes."""
+
+
+import numpy as np
+
+
+__all__ = ["convert_array_to", "validate_shape"]
+
+
+def convert_array_to(dtype):
+    """Return a function to convert arrays to the given type."""
+    def converter(array):
+        if array is None:
+            return None
+        return np.array(array, copy=False, dtype=dtype)
+    return converter
+
+
+# pylint: disable=too-many-branches
+def validate_shape(*shape):
+    """Return a validator for the shape of an array or the length of an iterable.
+
+    Parameters
+    ----------
+    shape
+        The expected shape. For general iterables, this should be just one
+        element. For arrays, also ``ndim`` is compared to the length of the
+        given shape.
+
+    Returns
+    -------
+    validator
+        A validator function for the attr library
+
+    Notes
+    -----
+    Each element of shape can be one of the following four things and should
+    match the shape of the attribute being validated:
+
+    1. An integer. In this case, this is the expected length along this axis.
+    2. None. In this case, the length of the array along this axis is not checked.
+    3. A string. In this case, it is the name of an integer attribute with the
+       expected length.
+    4. A tuple of a name and an integer. In this case, the name refers to another
+       attribute which is an array or some iterable. When the integer is 0,
+       just the length of the attribute is used.
+
+    """
+    def validator(obj, attribute, value):
+        # Build the expected shape, with the rules from the docstring.
+        expected_shape = []
+        for item in shape:
+            if isinstance(item, int) or item is None:
+                expected_shape.append(item)
+            elif isinstance(item, str):
+                expected_shape.append(getattr(obj, item))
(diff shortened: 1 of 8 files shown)


iodata/attrutils.py
```diff
@@ -0,0 +1,107 @@
...
+    return converter
+
+
+# pylint: disable=too-many-branches
+def validate_shape(*shape):
+    """Return a validator for the shape of an array or the length of an iterable.
+
+    Parameters
+    ----------
+    shape
```
FarnazH: It took me a little while to understand how this validator works. Is it possible to use a better argument name instead of `shape` and or improve the description?

iodata/test/test_basis.py
```diff
@@ -105,20 +105,39 @@ def test_shell_info_propertes():
     assert obasis.nbasis == 1 + 4 + 4 + 5 + 6 + 7 + 9
 
 
+def test_shell_validators():
+    Shell(0, [0, 0], ['c', 'c'], np.zeros(6), np.zeros((6, 2)))
```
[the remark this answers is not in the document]
tovrstra: I put it there because I expect it not to cause any error unlike the small variations on the following lines. It should add a comment to clarify this. A line could also be added to test explicitly that it passes the validation:

```python
# The following line is constructs a Shell instance with valid arguments.
# It should not raise a TypeError.
shell = Shell(0, [0, 0], ['c', 'c'], np.zeros(6), np.zeros((6, 2)))
# Rerun the validators as a double check.
attr.validate(shell)
```

See https://www.attrs.org/en/stable/api.html#attr.validate This would be double checking in the sense that the validators get already called in the constructor, but double checking doesn't hurt.

iodata/attrutils.py
tovrstra: Yep, that should definitely be done. Something like `shape_requirements` would already be better in the sense that this name makes the distinction from a regular `ndarray.shape` which only contains integers.

tovrstra:
@FarnazH Fixes for your comments are included.

I've also experimented a little further and learned that attr only runs its converters and validators when constructing a new object. (See new utit tests.) Converters and validators do not get triggered when assigning attributes. This means it remains easy to create objects with invalid attributes: first create it and then modify it. :frowning: 

To avoid this issue, the `frozen=True` option can be used:

```python
@attr.s(slots=True, frozen=True)
class Wacko:
    """Just a silly class for testing convert_array_to."""

    flip: np.ndarray = attr.ib(converter=convert_array_to(float))
```

This would fail the following test on the last line with an `attr.exceptions.FrozenInstanceError`:

```python
def test_convert_array_to_assign():
    wacko = Wacko(flip=None)
    assert wacko.flip is None
    wacko.flip = np.array([1.0, 3.0, -1.0])
```

That is a solution, but not a great one, because it also prohibits potentially legitimate code. That said, we had the same limitation with `NamedTuple`.

(I believe we ran into a similar issue before.)

tovrstra:
The lack of assignment validation seems to be an often-discussed issue of attr. There are some alternatives or extensions to consider:

- https://smarie.github.io/python-pyfields/
- https://github.com/pwwang/attr_property
- https://github.com/smarie/python-autoclass
- https://github.com/RussBaz/enforce
- https://github.com/Stewori/pytypes
- https://github.com/aldanor/typo
- https://github.com/agronholm/typeguard
- https://attrs-strict.readthedocs.io/en/latest/

Alternatively, one can also write a `__setattr__` method. The main catch seems to be the performance hit caused by the type checking. (on the order of microseconds) To avoid performance issues, `frozen=True` can be used in combination with `attr.evolve`.

tovrstra:
After reading and comparing some alternatives, it seems that the only suitable alternative to Attrs is pyfields (and related libraries like autoclass and vtypes, see https://github.com/smarie). Pyfields is not as polished and popular and it seems to have a slightly larger performance hit compared to attrs. Still, pyfields does trivially support validation upon attribute assignment. A comparable feature is being proposed by the `attrs` developer, see python-attrs/attrs#645, which would also address the concerns above.

The following would be a good temporary solution:

- Do not (yet) enforce validation upon attribute assignment.
- Perform (and facilitate) validation internally right before dumping data to a file or after loading it from a file. This can be implemented with https://www.attrs.org/en/stable/api.html#attr.validate

Main advantages:

- This can be implemented easily without adding or changing dependencies.
- Negligible impact on performance.
- No API change.
- Users are guaranteed to get consistent data when loading from a file, e.g. the bug fixed in this PR would be detected without adding extra tests.
- Dump functions can safely assume they receive valid `IOData` objects.
- Users can still write pythonic code in which an IOData object is temporarily invalid. It only needs to be valid when interacting with the iodata library, which makes life easier for end users. (This "advantage" could also cause confusion though.)

Limitations:

- The calls to [attr.validate](https://www.attrs.org/en/stable/api.html#attr.validate) could be redundant and do not automatically validate attributes of attributes. For that, we'll need to write a helper function. This is not such a big deal because the overhead is negligible.

As soon as `attr` supports validation upon assignment, we can still decide to start using it. The main disadvantage will be that the last advantage in the list above disappears.

tovrstra:
Superseded by #157.

PR #141, opened 2020-05-21
Add mkl dump functionality 

leila-pujal:
I added `dump_one` function for molekel file format and tested it similar to molden `dump_one` function.

### DIFF
--- a/iodata/formats/molekel.py
+++ b/iodata/formats/molekel.py
@@ -113,6 +113,7 @@ def _load_helper_obasis(lit: LineIterator) -> MolecularBasis:
 def _load_helper_coeffs(lit: LineIterator, nbasis: int) -> Tuple[np.ndarray, np.ndarray]:
     coeffs = []
     energies = []
+    irreps = []
 
     in_orb = 0
     for line in lit:
@@ -125,7 +126,8 @@ def _load_helper_coeffs(lit: LineIterator, nbasis: int) -> Tuple[np.ndarray, np.
             ncol = len(words)
             assert ncol > 0
             for word in words:
-                assert word == 'a1g'
+                # assert word == 'a1g'
+                irreps.append(word)
             cols = [np.zeros((nbasis, 1), float) for _ in range(ncol)]
             in_orb = 1
         elif in_orb == 1:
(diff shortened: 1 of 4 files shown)


tovrstra:
This looks already pretty good. Could you rename the functions `check_load_molden_dump_molekel_li2` and `check_load_fchk_dump_molekel_li2` to `test_load_molden_dump_molekel_li2` and `test_load_fchk_dump_molekel_li2`? If they are not meant to pass yet, you can always mark them with xfail: https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail Thanks.

leila-pujal:
`Check_load_molden_dump_molekel_li2` and `check_load_fchk_dump_molekel_li2` were changed to `test_load_molden_dump_molekel_li2` and `test_load_fchk_dump_molekel_li2`. In order to have these tests to pass, a new `compare_mols` function is defined in test_molekel.py, `compare_mols_diff_formats`.  It was not possible to use `compare_mols` function in common.py, as it is supposed to compare two IOData objects created from the same type of format file.  Also, some minor changes were made in molekel.py, related to atomic basis information for each center and saving/printing of irreps for orbital symmetry.

iodata/formats/molekel.py
```diff
@@ -112,7 +126,8 @@ def _load_helper_coeffs(lit: LineIterator, nbasis: int) -> Tuple[np.ndarray, np.
             ncol = len(words)
             assert ncol > 0
             for word in words:
-                assert word == 'a1g'
+                # assert word == 'a1g'
```
tovrstra, line 129: Feel free to remove this commented line. This assert should never have been here in the first place.

iodata/formats/molekel.py
```diff
@@ -223,13 +243,150 @@ def load_one(lit: LineIterator) -> dict:
...
+        f.write('\n')
+
+        # COEFF_BETA
+        f.write('$COEFF_BETA\n')
+        _dump_helper_coeffs(f, data, spin='b')
+
+        # OCC_BETA
+        f.write('$OCC_BETA\n')
+        _dump_helper_occ(f, data, spin='b')
+
```
tovrstra, line 339: If you want to deal with the generalized case, you might add something like

```python
elif data.mo.kind == 'generalized':
    raise ValueError("The MKL format does not support generalized orbitals..")
```

FarnazH: Thanks for your comment. Fixed it with an `else` statement to be more general.

iodata/formats/molekel.py
FarnazH: Removed. Thanks.

Issue #79, opened 2019-04-23
Add ordering from CCA to iodata.basis


tovrstra:
See following paper: https://onlinelibrary.wiley.com/doi/full/10.1002/jcc.20815 
"Components for integral evaluation in quantum chemistry" by Joseph P. Kenny  Curtis L. Janssen  Edward F. Valeev  Theresa L. Windus

PR #144, opened 2020-06-22
(WIP) Add JSON input/output functionality

wilhadams:
Add `load_one` and `dump_one` functions for the QCSchema JSON filetype.

As a result of inconsistent implementation in the QCSchema repo, QCElemental, and the Basis Set Exchange, doing this will require building an updated specification for the QCSchema. I've tried to keep this as backwards-compatible as possible, with some exceptions: QCElemental, for example, has a basis specification definition that differs from both other sources, however, that basis spec was never fully implemented in QCElemental, so there's no reason to be held hostage to that definition. 

There are currently 76 test files in the `json_tests_qcrecords` directory and 21 in the `json_tests_qcschemadev` that are being used to test most of the implementation, however, at this point there isn't a "perfect example" for every field in (any variant of) the schema. I will pick good examples from these batches and generate valid JSON files for cases without a test case. 

This PR is still a bit of a mess, I'll have it cleaned up soon.


wilhadams:
QCSchema uses CCA conventions, so this is relevant to #79

PaulWAyers:
Does this mean that Issue #79 will be resolved after this pull request is merged?

wilhadams:
@PaulWAyers that was vague, sorry. Yes, I'll make sure that this PR closes #79

wilhadams:
Different parts of the QCSchema are at different levels of maturity right now, in addition to different levels of availability for examples, ranging from none to not many. I'm going to close this PR and break it into several smaller PRs to get the better-tested stuff out and start working on generating tests for the other parts.

PR #149, opened 2020-06-25
Add `load_one` for Q-Chem log file

fwmeng88:
Implementation includes
- [x] Lower level parser for Q-Chem log file, 509ba22
- [x] Higher level `load_one` function for Q-Chem log file, 33d2a25
- [x] Tests for Lower level parser and higher level `load_one` function, f8a791a and 9622008
- [x] Molecular orbital object for Q-Chem return
- [x] Fix pycodestyle errors 7e522b3
- [x] Add tests for molecular object opbject in 4132217

### DIFF
--- a/iodata/formats/qchemlog.py
+++ b/iodata/formats/qchemlog.py
@@ -84,7 +84,7 @@ def load_one(lit: LineIterator) -> dict:
     moments_labels = ['dipole_moment', 'quadrupole_moments', 'dipole_tol']
     moments = {label: data.get(label, None) for label in moments_labels}
     # extra information
-    extra_labels = ['spin_multi', 'nuclear_repulsion_energy', 'nbasis', 'charge',
+    extra_labels = ['spin_multi', 'nuclear_repulsion_energy', 'charge',
                     'polarizability_tensor', 'hessian',
                     'imaginary_freq', 'vib_energy']
     extra = {label: data.get(label, None) for label in extra_labels}
--- a/iodata/formats/qchemlog.py
+++ b/iodata/formats/qchemlog.py
@@ -18,7 +18,10 @@
 # --
 """Q-Chem Log file format.
 
+This module will load Q-Chem log file into IODATA.
 """
+
+
 import re
 from typing import List, Tuple
 
@@ -39,7 +42,7 @@ PATTERNS = ['*.qchemlog']
 def load_one(lit: LineIterator) -> dict:
     """Do not edit this docstring. It will be overwritten."""
     data = load_qchemlog_low(lit)
-    result_labels = ['atcoords', 'atmasses', 'atnums','charge',
+    result_labels = ['atcoords', 'atmasses', 'atnums', 'charge',
                      'charge', 'energy', 'g_rot', 'run_type']
     result = {label: data.get(label, None) for label in result_labels}
     # build molecular orbitals
@@ -99,21 +102,22 @@ def load_qchemlog_low(lit: LineIterator) -> dict:
         # get the atomic information
         if line.startswith('$molecule'):
             data['charge'], data['spin_multi'], data['natom'], \
-              data['atnums'], data['atcoords'] = _helper_atoms(lit)
+                data['atnums'], data['atcoords'] = _helper_atoms(lit)
             natoms = len(data['atnums'])
         # job specifications
         elif line.startswith('$rem'):
             data['run_type'], data['method'], data['basis_set'], \
-              data['unrestricted'], data['symm'] = _helper_job(lit)
+                data['unrestricted'], data['symm'] = _helper_job(lit)
         # standard nuclear orientation
         elif line.startswith('Standard Nuclear Orientation (Angstroms)'):
             # atnums, alpha_elec, beta_elec, nbasis, nuclear_replusion_energy, energy, atcoords
             _, data['alpha_elec'], data['beta_elec'], data['nbasis'], \
-              data['nuclear_repulsion_energy'], data['energy'], _ = _helper_electron(lit)
+                data['nuclear_repulsion_energy'], data['energy'], _ = _helper_electron(lit)
         # orbital energies
         elif line.startswith('Orbital Energies (a.u.)'):
             data['alpha_mo_occupied'], data['beta_mo_occupied'], data['alpha_mo_unoccupied'], \
-              data['beta_mo_unoccupied'], data['norba'], data['norbb'] = _helper_orbital_energies(lit)
+                data['beta_mo_unoccupied'], data['norba'], \
+                data['norbb'] = _helper_orbital_energies(lit)
         # mulliken charges
         elif line.startswith('Ground-State Mulliken Net Atomic Charges'):
             data['mulliken_charges'] = _helper_mulliken(lit)
@@ -230,10 +234,10 @@ def _helper_orbital_energies(lit: LineIterator) -> Tuple:
     norbb = len(beta_mo_occupied + beta_mo_unoccupied)
     # todo: not sure how to arrange the four type of molecular orbital energies here
     return np.array(alpha_mo_occupied, dtype=np.float), \
-           np.array(beta_mo_occupied, dtype=np.float), \
-           np.array(alpha_mo_unoccupied, dtype=np.float), \
-           np.array(beta_mo_unoccupied, dtype=np.float), \
-           norba, norbb
+        np.array(beta_mo_occupied, dtype=np.float), \
+        np.array(alpha_mo_unoccupied, dtype=np.float), \
+        np.array(beta_mo_unoccupied, dtype=np.float), \
+        norba, norbb
 
 
 def _helper_section(start: str, end: str, lit: LineIterator, backward: bool = False) -> List:
@@ -285,7 +289,7 @@ def _helper_dipole_moments(lit: LineIterator) -> Tuple:
     quadrupole_moments.extend(next(lit).strip().split())
     quadrupole_moments = [dipole for idx, dipole in enumerate(quadrupole_moments) if idx % 2 != 0]
     return np.array(dipole_moment, dtype=np.float), \
-           np.array(quadrupole_moments, dtype=np.float), dipole_tol
+        np.array(quadrupole_moments, dtype=np.float), dipole_tol
 
 
 def _helper_polar(lit: LineIterator) -> np.ndarray:
@@ -304,7 +308,7 @@ def _helper_hessian(lit: LineIterator, natoms: int) -> np.ndarray:
     """Load hessian matrix."""
     # hessian in Cartesian coordinates, shape(3 * natom, 3 * natom)
     col_idx = [int(i) for i in next(lit).strip().split()]
-    hessian = np.empty((natoms*3, natoms*3), dtype=object)
+    hessian = np.empty((natoms * 3, natoms * 3), dtype=object)
     for line in lit:
         if line.strip().startswith('****************'):
             break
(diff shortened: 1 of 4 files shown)


iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,370 @@
...
+    result = {label: data.get(label, None) for label in result_labels}
+    # build molecular orbitals
+    # todo: double check if this is right
+    # mo_energies
+    # # restricted case
+    # if not data['unrestricted']:
+    #     mo_energies = np.concatenate(
+    #         (data['alpha_mo_occupied'], data['alpha_mo_unoccupied']), axis=0)
+    #     mo_coeffs = np.empty((data['nbasis'], data['norba'])) * np.nan
+    #     mo_occs = np.zeros(mo_coeffs.shape[1]) * np.nan
```
FarnazH, line 56: It seems that MO `coeffs` and `occs` are empty/zeros arrays; is it what you use to compute the number of electrons? If not, can you please explain what you get? The `coeffs` and `occs` don't exist in this Q-Chem log file.

fwmeng88: Yeah, `coeffs` and `occs` don't exist in qchem log file. I didn't use them to compute the number of electrons but adding number of alpha electrons and beta electrons, https://github.com/fwmeng88/iodata/blob/f0631ff3b76212877561da85aa48937afc08f978/iodata/formats/qchemlog.py#L68. 

The problem here is if I call `mol.charge` it will raise the wrong charge number with the above section uncommented.
```python
========================== test session starts ===========================
platform linux -- Python 3.7.7, pytest-5.0.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/legend/Desktop/work/iodata_project/iodata
collected 2 items                                                        

iodata/test/test_qchemlog.py .F                                    [100%]

================================ FAILURES ================================
_________________________ test_load_one_qchemlog _________________________

    def test_load_one_qchemlog():
        with path('iodata.test.data', 'water_hf_ccpvtz_freq_qchem.out') as fn_qchemlog:
            mol = load_one(str(fn_qchemlog), fmt='qchemlog')
>       assert mol.charge == 0
E       AssertionError: assert nan == 0
E        +  where nan = IOData(atcharges={}, atcoords=array([[ 0.00575,  0.00426, -0.00301],\n       [ 0.27588,  0.88612,  0.25191],\n       [ 0... obasis_name='cc-pvtz', one_ints={}, one_rdms={}, run_type='freq', _spinpol=None, title=None, two_ints={}, two_rdms={}).charge

iodata/test/test_qchemlog.py:139: AssertionError
=================== 1 failed, 1 passed in 0.21 seconds ===================

```
Thank you.

FarnazH: @fwmeng88, the test passes for me (and on Travis-CI as well). So, I cannot reproduce your problem.

FarnazH:
When an instance of `MolecularOrbitals` is given to `IOData`, the sum of MO occupations is used to calculate the number of electrons. So, if `mo.occs` is `np.nan`, you get a `nan` for the number of electrons/charges. Even though occupations are not printed in the Q-Chem log file, you can use number of alpha & beta electrons to assign proper `mo.occs`, which is done in cd5e515.

iodata/formats/qchemlog.py
fwmeng88: Problem fixed in cd5e515.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,366 @@
...
+                    'mulliken_charges', 'polarizability_tensor', 'hessian',
+                    'imaginary_freq', 'vib_energy', 'entropy_dict']
+    extra = {label: data.get(label, None) for label in extra_labels}
+    # unit conversions for vibrational energy
+    extra['vib_energy'] = extra.get('vib_energy') * kcalmol
+    # convert kcal/mol to atomic units
+    enthalpy_dict = {k: v * kcalmol for k, v in data['enthalpy_dict'].items()}
+    # todo: unit conversions for entropy
+    extra['enthalpy_dict'] = enthalpy_dict
+    extra['moments'] = moments
```
FarnazH, line 96: The moments should be properly stored in `result["moments"]` dictionary, see https://github.com/theochem/iodata/blob/master/iodata/iodata.py#L140

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,366 @@
...
+from ..periodic import sym2num
+from ..utils import LineIterator, kcalmol
+
+__all__ = []
+
+PATTERNS = ['*.qchemlog']
+
+
+@document_load_one("qchemlog", ['atcoords', 'athessian', 'atmasses', 'atnums', 'charge', 'energy',
+                                'g_rot', 'mo', 'lot', 'nelec', 'obasis_name', 'run_type', 'extra'])
```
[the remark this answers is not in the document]
fwmeng88: Fixed in c214bfe.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,366 @@
...
+    extra_labels = ['spin_multi', 'nuclear_repulsion_energy', 'nbasis', 'charge',
+                    'mulliken_charges', 'polarizability_tensor', 'hessian',
+                    'imaginary_freq', 'vib_energy', 'entropy_dict']
+    extra = {label: data.get(label, None) for label in extra_labels}
+    # unit conversions for vibrational energy
+    extra['vib_energy'] = extra.get('vib_energy') * kcalmol
+    # convert kcal/mol to atomic units
+    enthalpy_dict = {k: v * kcalmol for k, v in data['enthalpy_dict'].items()}
+    # todo: unit conversions for entropy
+    extra['enthalpy_dict'] = enthalpy_dict
```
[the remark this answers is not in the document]
fwmeng88: It's already in atomic units, so no further coding is needed. But still thanks for noticing the unit conversion issue.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,366 @@
...
+    result['nelec'] = data['alpha_elec'] + data['beta_elec']
+    result['obasis_name'] = data['basis_set'].lower()
+    # moments
+    moments_labels = ['dipole_moment', 'quadrupole_moments', 'dipole_tol']
+    moments = {label: data.get(label, None) for label in moments_labels}
+    # extra information
+    extra_labels = ['spin_multi', 'nuclear_repulsion_energy', 'nbasis', 'charge',
+                    'mulliken_charges', 'polarizability_tensor', 'hessian',
+                    'imaginary_freq', 'vib_energy', 'entropy_dict']
+    extra = {label: data.get(label, None) for label in extra_labels}
```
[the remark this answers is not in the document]
fwmeng88: Thanks for helping to improve the codes!

1. "mulliken_charges" now is stored as `result["atcharges"]`, 070a0af
2. "nbasis" is removed as it's duplicating with `MolecularOrbitals` instance, fb5ed59
3.  "charge" is stored in `result["charge"]`, 27d23f5
4. "hessian" is stored in `result["athessian"] = data.get("hessian")`, b61cc46

iodata/formats/qchemlog.py
fwmeng88: Thanks for noticing this. They have been fixed in 1bb64ee and 16071c7.

iodata/formats/qchemlog.py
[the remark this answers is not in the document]
FarnazH: My bad. I didn't notice the unit conversion in the previous lines.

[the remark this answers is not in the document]
fwmeng88: Never mind. It's good to raise the point and we should be careful with the units.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,368 @@
...
+    if 'vib_energy' in extra:
+        extra['vib_energy'] *= kcalmol
+
+    # if present, convert enthalpy terms from kcal/mol to "atomic units + K"
+    if 'enthalpy_dict' in data:
+        extra['enthalpy_dict'] = {k: v * kcalmol for k, v in data['enthalpy_dict'].items()}
+
+    # if present, convert entropy terms from cal/mol.K to "atomic units + Kalvin"
+    if 'entropy_dict' in data:
+        extra['entropy_dict'] = {k: v * calmol for k, v in data['entropy_dict'].items()}
```
FarnazH, line 110: @tovrstra what are your thoughts on storing thermodynamic properties cleanly? I am worried that `extra` is becoming a place to dump things that are not clearly defined... Here, it makes sense to probably parse temperature and pressure as well, but then you add more items to the dictionary.

@wilhadams is there any specific convention for storing thermodynamic properties in Schema?

wilhadams: @FarnazH currently there's no convention.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,368 @@
...
+    # mulliken charges
+    if data.get("mulliken_charges") is not None:
+        result["atcharges"] = {"mulliken": data["mulliken_charges"]}
+
+    # build molecular orbitals
+    # ------------------------
+    # restricted case
+    if not data['unrestricted']:
+        mo_energies = np.concatenate((data['mo_a_occ'], data['mo_a_vir']), axis=0)
+        mo_coeffs = np.empty((data['nbasis'], data['norba'])) * np.nan
```
tovrstra, line 63: Would there be a way to avoid creating a large matrix with nans? E.g. would the following be ok?
```suggestion
        mo_coeffs = np.empty((0, data['norba']))
```
If not, it would be good for readability to use [np.full](https://numpy.org/doc/stable/reference/generated/numpy.full.html) instead of multiplying with nan.

iodata/formats/qchemlog.py
```diff
@@ -0,0 +1,368 @@
...
+        mo_occs = np.zeros(mo_coeffs.shape[1])
+        mo_occs[:data['alpha_elec']] = 1.0
+        mo_occs[:data['beta_elec']] += 1.0
+        mo = MolecularOrbitals("restricted", data['norba'], data['norba'],
+                               mo_occs, mo_coeffs, mo_energies, None)
+    # unrestricted case
+    else:
+        mo_energies = np.concatenate((data['mo_a_occ'], data['mo_a_vir'],
+                                      data['mo_b_occ'], data['mo_b_vir']), axis=0)
+        mo_coeffs = np.empty((data['nbasis'], data['norba'] + data['norbb'])) * np.nan
```
tovrstra, line 73: same comment as above.

iodata/formats/qchemlog.py
tovrstra: I would not mind adding these to the extra dictionary for now. It might be useful to document that the keys in extra are never to be considered API stable. This gives us the freedom to migrate some commonly used items in `extra` to more final attributes of `IOData`. @FarnazH: would that make sense?

iodata/formats/qchemlog.py
fwmeng88: Thanks for noticing this, Toon! I think it would be great to keep the right shape, instead of return an empty array with shape `(0, data['norba'])`. We can try 

```python
np.full((data['nbasis'], data['norba']), np.nan)
```

as you suggested, which is cleaner and faster. 
Would this be good? @tovrstra

tovrstra: OK, Good for me.

We should address this in future though, e.g. by allowing `mo_coeffs=None` when creating a `MolecularOrbitals` instance. This would be more in line with the data (not) present in the qhem log file and it would be efficient in terms of memory. I've opened an issue for it, which we can discuss later. (#168)  This way, we can merge your PR more quickly.

iodata/formats/qchemlog.py
fwmeng88: Fixed in 45eeba0. Thanks for pointing it out.

iodata/formats/qchemlog.py
fwmeng88: This make more sense. I have fixed the codes with `np.full()` which is more efficient, 45eeba0.

We can discuss more in GitHub issues and thanks for giving this nice point!

iodata/formats/qchemlog.py
FarnazH: Make sense @tovrstra. I will update the `extra` docstring in a future PR.

PR #153, opened 2020-07-05
Extension to orcalog format to read in enegies from the scf cycle

evohringer:
A small extension which stores the energies of the scf cycle in the extra dictionary as "scf_energies" of the IOData class.
(It will only keep the energies of the last cycle)


iodata/test/test_orcalog.py
```diff
@@ -44,7 +44,10 @@ def test_load_water_number():
         mol.atcoords[0] - mol.atcoords[2]) / angstrom, 0.9500, atol=1.e-5)
     assert_allclose(np.linalg.norm(
         mol.atcoords[1] - mol.atcoords[2]) / angstrom, 1.5513, atol=1.e-4)
+    # check energies of scf cycles
+    energies = np.array([-76.34739931, -76.34740001, -76.34740005, -76.34740029])
+    assert_allclose(mol.extra['scf_energies'], energies)
     # check scf energy
-    assert_allclose(mol.energy, -74.959292304818, atol=1e-8)
+    assert_allclose(mol.energy, -76.347791524303, atol=1e-8)
```
FarnazH, line 51: The energy values (total/SCF) are parsed properly, but a naive question about the ORCA energy values. I am confused about why the total energy doesn't match the energy of the last iteration. Checking the log file, the total energy corresponds to the 4th SCF cycle.

iodata/formats/orcalog.py
```diff
@@ -33,7 +33,7 @@
 PATTERNS = ['*.out']
 
 
-@document_load_one("Orca output", ['atcoords', 'atnums', 'energy', 'moments'])
+@document_load_one("Orca output", ['atcoords', 'atnums', 'energy', 'moments', 'extra'])
```
FarnazH, line 36: Just to make sure, is the "SCF ITERATIONS" always printed by default by ORCA? If not, I think, the "extra" should be listed as the "ifpresent" argument...

iodata/test/test_orcalog.py
evohringer: I think the reason is that ORCA uses a finer grid for the final evaluation of the energy than for the SCF iterations.

iodata/formats/orcalog.py
evohringer: I think by default they are always printed out.

iodata/formats/orcalog.py
```diff
@@ -50,6 +50,10 @@ def load_one(lit: LineIterator) -> dict:
         # to maintain the ones from the final SCF iteration in e.g. optimization run
         if line.startswith('CARTESIAN COORDINATES (A.U.)'):
             result['atnums'], result['atcoords'] = _helper_geometry(lit, natom)
+        # Read the energies of each SCF cycle in iodata.extra
+        if line.startswith('ITER'):
```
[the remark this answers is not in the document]
evohringer: Good point. I changed it to SCF ITERATIONS. See new commit. Thanks

iodata/test/test_orcalog.py
tovrstra: It could also be for technical reasons, e.g. one may compute a cheaper approximation of the energy which is sufficient for just checking SCF convergence. E.g. the sum of occupied orbitals energies plus the usual correction can be cheaper but is not exact when the orbitals have not fully converged.

PR #154, opened 2020-07-05
Docstring Edits

FarnazH:
This includes @PaulWAyers's edits to docstrings.

### DIFF
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -137,15 +137,17 @@ class Shell(NamedTuple):
 
 
 class MolecularBasis(NamedTuple):
-    """Describe a complete molecular orbital or density basis set.
+    """A complete molecular orbital or density basis set.
 
     Attributes
     ----------
     shells
-        a list of objects of the type Shell
+        A list of objects of type Shell which can support generalized contractions.
     conventions
-        a dictionary with as key a typle of angular momentum integer and kind
-        character, and as value a list of basis function strings, e.g.
+        A dictionary specifying the ordered basis functions for a given angular momentum and kind.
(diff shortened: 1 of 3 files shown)


FarnazH:
@tovrstra, the failed tests are regarding 4 instances of "pydocstyle-D401: First line should be in imperative mood; try rephrasing". Personally I don't like phrasing the attribute documentation starting with a verb (like "Return ..."), but can understand why python thinks of it as a function... Let me know what is your preference, and I can change things. Thanks.

tovrstra:
@Farnaz I'm fine with disabling that check. There are quite a few cases where starting with a verb is not desirable. It is a good habit when it makes sense but enforcing it seems a little strict.

tovrstra:
P.S. You can change this by adding d401 to ~/.pydocstylerc: https://github.com/theochem/iodata/blob/master/.pydocstylerc

(See http://www.pydocstyle.org/en/5.0.1/error_codes.html for documentation.)

FarnazH:
@tovrstra, do we want to disable it for all modules? I think it is good to start with an imperative mood for functions/methods, so it is a good check. I was thinking that we can just ignore Travis and proceed with the merge based on the reviewer's discretion.

tovrstra:
@FarnazH I'd like to avoid this to keep final tests on the master branch overseeable when making a release. 
Would local exclusion on the four lines be an option? (See http://www.pydocstyle.org/en/5.0.2/usage.html#in-file-configuration)

FarnazH:
@tovrstra, thanks. That was a great alternative. I disabled it using pydocstyle in-line command.

PR #157, opened 2020-07-09
Convert all remaining namedtuples to attr classes

### LINKED ISSUE #201: Document required `attrs` package
The `attrs` package is requred for IOData, but there's no mention of it anywhere in the README nor the developer documentation. This is also important because there's also an `attr` package, with similar functionality, and the import statement for the `attrs` package is `import attr`, which is confusing. I think putting a brief overview of how `attrs` works within IOData into the developer docs would make it easier for new developers to contribute.

tovrstra:
Fixes #201 amongst other things:
- Improved attribute validation, both in constructor and upon attribute assignment (requires latest attrs version).
- `attrutil` module to facilitate writing validators.
- Fixed bug in CP2K loader.
- Document attr behavior in `CONTRIBUTING.rst`
- Few other fixes needed to satisfy validators.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -325,7 +325,42 @@ to avoid duplicate efforts.
     results in minor corrections at worst. We'll do our best to avoid larger
     problems in step 1.
 
+
+Notes on attrs
+--------------
+
+IOData uses the `attrs`_ library, not to be confused with the `attr`_ library,
+for classes representing data loaded from files: ``IOData``, ``MolecularBasis``,
+``Shell``, ``MolecularOrbitals`` and ``Cube``. This enables basic attribute
+validation, which eliminates potentially silly bugs. The following two tricks
+might be convenient with working with these classes:
+
+- The data can be turned into plain Python data types with the ``attr.asdict``
+  function. Make sure you add the ``retain_collection_types=True`` option, to
+  avoid the following issue: https://github.com/python-attrs/attrs/issues/646
+  For example.
+
+  .. code-block:: python
+
+      from iodata import load_one
+      import attr
+      iodata = load_one("example.xyz")
+      fields = attr.asdict(iodata, retain_collection_types=True)
+
+- A shallow copy with a few modified attributes can be created with the evolve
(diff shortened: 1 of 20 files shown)


tovrstra:
The motivation for the changes in this PR are outlined in more detail in #138. Most changes from that PR were also used here, but with some modifications and corrections.

tovrstra:
@FarnazH I removed the code again that explicitly runs the validators before dumping and after loading. Even without these, just by turning `Shell` into an attr class and adding validators, the bug in `cp2klog.py` already raises an error when constructing the `Shell` instance. Hence, running the validators once more seemed overkill. As soon as tests pass, it is ready for review.

tovrstra:
@FarnazH One thing we should consider adding here is to update all NamedTuple classes into attr classes. This may resolve issues like the one you mentioned here: https://github.com/theochem/iodata/pull/188#discussion_r474336204

tovrstra:
@wilhadams This PR switches all namedtuples to attr classes, adding more validation and a few minor bug fixes. Does the last section in `CONTRIBUTING.rst` address your concern?

CONTRIBUTING.rst
```diff
@@ -325,7 +325,42 @@ to avoid duplicate efforts.
     results in minor corrections at worst. We'll do our best to avoid larger
     problems in step 1.
 
+
+Notes on attrs
```
wilhadams, line 329: This is really helpful, and exactly what I was looking for!

wilhadams (approved):

CONTRIBUTING.rst
tovrstra: Thanks! I'm going to update the contributing doc a bit more. It turns out that the new attr release from 9 days ago has few extra tricks, so I'll see if there is anything worth mentioning.

tovrstra: @wilhadams I'm going to rerequest review since I've now updated the code and documentation to use a few nice features of attrs 20.1.0, which just got released.

tovrstra:
I'm going to close this PR and reopen it because most of the comments above are outdated.

PR #160, opened 2020-07-15
Add .WFN dump_one()

msricher:
I copied Leila's code for decontracting and denormalizing the MO coefficients.

I still need to add tests.  I tried copying some stuff from Leila's test file for WFX, but none of her test functions seem to work yet?

I'll look at testing this once I have something nice to copy. But here is the code, anyway!

@FarnazH


leila-pujal:
Tests in test_wfx.py in wfx_dump branch from my IOData repo or fwmeng88 should work...I think you copied from the pull request, which is from fwmeng88 wfx_dump branch so that is updated. Could you post some output from pytest? maybe I forgot to commit something and that's why they don't work...Thanks!

msricher:
> Tests in test_wfx.py in wfx_dump branch from my IOData repo or fwmeng88 should work...I think you copied from the pull request, which is from fwmeng88 wfx_dump branch so that is updated. Could you post some output from pytest? maybe I forgot to commit something and that's why they don't work...Thanks!

I wasn't using Pytest, that's my problem.  I also didn't recognize the Pytest syntax for using the `tmpdir`.  I'll try that out.  Thank you!

msricher:
I am not sure why these tests fail.  I checked the files myself, and I checked the values compared by compare_mol, i.e. charge1 and charge2, and they are the same.

PaulWAyers:
Re: 9f254a8
Are we sure we can't use F12.6? Some people (and original AIMPAC wfn) use F12.8 but other codes use F12.6. I'd prefer F12.6 as otherwise orbital energies that are less than -99.99999999 cannot be reported. And one very quickly finds orbital energies smaller than this (for 1s core orbitals as one moves down the periodic table, roughly starting with Phosphorous Z=15). However, with F12.6, one can go as far as -9999.999999, which gets us roughly to Z=100. And we don't really need (or want) 8 digits of precision in orbital energies anyway.

msricher:
> Re: [9f254a8](https://github.com/theochem/iodata/commit/9f254a8e4973045f4e60dbb4dd9c6c78aeb35c9d)
> Are we sure we can't use F12.6? Some people (and original AIMPAC wfn) use F12.8 but other codes use F12.6. I'd prefer F12.6 as otherwise orbital energies that are less than -99.99999999 cannot be reported. And one very quickly finds orbital energies smaller than this (for 1s core orbitals as one moves down the periodic table, roughly starting with Phosphorous Z=15). However, with F12.6, one can go as far as -9999.999999, which gets us roughly to Z=100. And we don't really need (or want) 8 digits of precision in orbital energies anyway.

OK, I can make it 12.6, but then the tests don't pass because some of the test .wfn files use 12.8, and if I use 12.6 precision, some of the tests fail.  So either we use 12.8 or Farnaz can make the `compare_mols` function less strict?

iodata/test/test_wfn.py
```diff
@@ -306,3 +308,88 @@ def test_load_one_cah110_hf_sto3g_g09():
...
+    check_load_dump_consistency('o2_uhf.wfn', tmpdir)
+    check_load_dump_consistency('o2_uhf_virtual.wfn', tmpdir)
+
+
+def test_load_dump_consistency_from_fchk_h2o(tmpdir):
+    check_load_dump_consistency('h2o_sto3g.fchk', tmpdir, fmt_from='fchk', fmt_to='wfn')
+
+
+def test_load_dump_consistency_from_molden_nh3(tmpdir):
+    check_load_dump_consistency('nh3_molden_cart.molden', tmpdir, fmt_from='molden', fmt_to='wfn')
```
FarnazH, line 395: @msricher can you please check why these tests fail?

iodata/test/test_wfn.py
```diff
@@ -306,3 +308,88 @@ def test_load_one_cah110_hf_sto3g_g09():
...
+    check_load_dump_consistency('lih_cation_rohf.wfn', tmpdir)
+
+
+def test_load_dump_consistency_h2o(tmpdir):
+    check_load_dump_consistency('h2o_sto3g.wfn', tmpdir)
+    check_load_dump_consistency('h2o_sto3g_decontracted.wfn', tmpdir)
+
+
+def test_load_dump_consistency_lif(tmpdir):
+    check_load_dump_consistency('lif_fci.wfn', tmpdir)
```
FarnazH, line 357: @msricher can you please check why this test fails? It has fractional orbital occupations.

iodata/test/test_wfn.py
```diff
@@ -78,7 +79,7 @@ def test_load_wfn_low_h2o():
     data = helper_load_wfn_low('h2o_sto3g.wfn')
     # unpack data
     title, atnums, atcoords, centers, type_assignments = data[:5]
-    exponents, mo_count, occ_num, mo_energy, coefficients, energy, _ = data[5:]
+    exponents, mo_count, occ_num, mo_energy, coefficients, energy, _, _ = data[5:]
```
tovrstra, line 82: @msricher Would it make sense to add unit tests for the last two variables? They are now underscores but they correspond to `virial` and `mo_spin`, I believe.

msricher: I don't know.  There was already one underscore when I started working on this.  And nobody wanted me to add more test files with the MOSPIN section.

tovrstra: I'm not familiar with the MOSPIN section, so I can't comment on that. Would it still be useful to check the virial then?

msricher: Yeah, it would be.  Sorry, my wrists are hurting today, I need to take a break.  So I can't do it for a few days...

tovrstra: No problem and no rush either. Take all the time you need.

msricher: OK, I added the tests for the virial.

tovrstra: Thanks!

leila-pujal:
Yes, I tested to load several wfn files, with AIMALL Version 14.06.21 Professional, for water with the different formats we’ve been checking (wfx, wfn, molden and fchk) for sto-3g and cc-pvtz basis set. As far as I checked I didn’t notice any error message and information I checked was correct (number of primitives, QTAIM critical points analysis and net charge).

FarnazH:
Thanks a lot for your due diligence @leila-pujal. Do you have any further comments @tovrstra?

msricher:
Ok, thanks Leila!  I also tested the MOSPIN stuff manually (although not
super thoroughly), for what it's worth.  So I think we just wait for Toon's
comments.

PR #164, opened 2020-07-16
Dump fchk

rayhe88:
The fump_one and dump_many functions were programmed, for the creation of the **FCHK** files, some fields are not present in the **IOData** object, so those were not dumped. Take this information for reference http://theochem.mercer.edu/chm295/g09ur/f_formchk.htm


iodata/formats/fchk.py
```diff
@@ -435,3 +439,225 @@ def _triangle_to_dense(triangle: np.ndarray) -> np.ndarray:
...
+        items[0] = "SP"
+    print(f"{items[0].upper():10s}{items[1].upper():30s}{items[2].upper():>33s}", file=f)
+
+    # write basic information
+    _dump_integer_scalars("Number of atoms", data.natom, f)
+    _dump_integer_scalars("Number of electrons", int(data.nelec), f)
+    if data.charge is not None:
+        _dump_integer_scalars("Charge", int(data.charge), f)
+    if data.mo is not None:
+        # assign number of alpha and beta electrons
```
tovrstra, line 509: Could you add a few checks for the occupation numbers?

- FCHK only supports integer occupation numbers. (It cannot be used to store natural orbitals with fractional occupation numbers.)
- FCHK assumes occupied orbitals come first, followed by virtual, which may not be the case in the IOData object.

In both cases, if the IOData is not compatible with the assumptions of the FCHK format, an exception should be raised.

In general, we need a good mechanism form "preparing" an IOData object if it can be adapted to become compatible with the assumptions of the file format. E.g. in the case of natural orbitals, the FCHK file can still contain the 1RDM. Adaptations to the IOData object are also needed in the WFN file, e.g. to decontract the basis. To keep the length of the `dump_one` and `dump_many` functions under control, it would be good to factor out such manipulations (and possibly introduce a file format API for it). @FarnazH @PaulWAyers @evohringer: we've discussed this before and I would prefer to make an issue for this. It is not something that should be handled in this PR (or stop it from merging).

iodata/formats/fchk.py
```diff
@@ -435,3 +439,225 @@ def _triangle_to_dense(triangle: np.ndarray) -> np.ndarray:
...
+            if k == 5 or i == nval - 1:
+                print("", file=f)
+                k = 0
+
+
+@document_dump_one(
+    "Gaussian Formatted Checkpoint",
+    ['atcoords', 'atnums', 'energy', 'mo', 'obasis'],
+    ['atcharges', 'atcorenums', 'atfrozen', 'atgradient', 'athessian', 'atmasses', 'lot',
+     'one_rdms', 'obasis_name', 'extra', 'moments'])
```
tovrstra, line 491: The required and optional `IOData` attributes need to be updated to reflect the behavior of the code. I believe it should be as follows, but please double check:
```suggestion
    ['atnums', 'atcorenums'],
    ['atcharges', 'atcoords', 'atfrozen', 'atgradient', 'athessian', 'atmasses', 
￼     'charge', 'energy', 'lot', 'mo', 'one_rdms', 'obasis', 'obasis_name',
     'extra', 'moments'])
```

iodata/formats/fchk.py
```diff
@@ -435,3 +439,225 @@ def _triangle_to_dense(triangle: np.ndarray) -> np.ndarray:
...
+
+@document_dump_many(
+    "Gaussian Formatted Checkpoint",
+    ['atcoords', 'atnums', 'energy', 'mo', 'obasis'],
+    ['atcharges', 'atcorenums', 'atfrozen', 'atgradient', 'athessian', 'atmasses', 'lot',
+     'one_rdms', 'obasis_name', 'extra', 'moments'])
+def dump_many(f: TextIO, datas: Iterator[IOData]):
+    """Do not edit this docstring. It will be overwritten."""
+    for data in datas:
+        dump_one(f, data)
```
tovrstra, line 663: This will not be compatible with `load_many`, which does not read a concatenation of FCHK files. It was specifically written to extract IRCs and (relaxed) scans from a single FCHK file as a list of IOData objects. I would suggest to ommit this function from the PR and open a separate PR with a `dump_many` that is compatible with `load_many`.

iodata/test/test_fchk.py
```diff
@@ -514,3 +515,184 @@ def test_load_nbasis_indep(tmpdir):
...
+
+
+# def test_dump_fchk_from_wfn_fci_lih_cation(tmpdir):
+#     check_load_dump_consistency(tmpdir, 'lih_cation_fci.wfn')
+
+
+# def test_dump_fchk_from_wfn_fci_lif(tmpdir):
+#     check_load_dump_consistency(tmpdir, 'lif_fci.wfn')
+
+
```
tovrstra, line 620: It would be useful to include these tests. Why were they comment out? If there is a pending issue with them, please add pytest markers.

```suggestion
@pytest.mark.xfail
def test_dump_fchk_from_wfn_fci_lih_cation(tmpdir):
    check_load_dump_consistency(tmpdir, 'lih_cation_fci.wfn')


@pytest.mark.xfail
def test_dump_fchk_from_wfn_fci_lif(tmpdir):
    check_load_dump_consistency(tmpdir, 'lif_fci.wfn')


```

iodata/test/test_fchk.py
```diff
@@ -514,3 +515,184 @@ def test_load_nbasis_indep(tmpdir):
...
+    check_load_dump_consistency(tmpdir, 'lih_cation_uhf.wfx')
+    check_load_dump_consistency(tmpdir, 'lih_cation_rohf.wfx')
+
+
+# def test_dump_fchk_from_wfx_lih_cisd_cation(tmpdir):
+#     check_load_dump_consistency(tmpdir, 'lih_cation_cisd.wfx')
+
+
+# def test_dump_fchk_from_wfx_cah110(tmpdir):
+#     check_load_dump_consistency(tmpdir, 'cah110_hf_sto3g_g09.wfx')
```
tovrstra, line 661: Same comment as above.

iodata/test/test_fchk.py
```diff
@@ -514,3 +515,184 @@ def test_load_nbasis_indep(tmpdir):
...
+    check_load_dump_consistency(tmpdir, "nitrogen-mp3.fchk")
+
+
+def test_dump_many_consistency(tmpdir):
+    trj = load_fchk_trj_helper("peroxide_opt.fchk")
+    # Write in temporary folder and then read it
+    for mol in trj:
+        mol.nelec = 18
+    fn_tmp = os.path.join(tmpdir, 'test')
+    dump_many(trj, fn_tmp, fmt='fchk')
```
tovrstra, line 698: In line with the comment above on `dump_many`, I would also suggest to remove this test.

iodata/formats/fchk.py
```diff
@@ -170,8 +172,9 @@ def load_one(lit: LineIterator) -> dict:
     # D) Load the wavefunction
     # Handle small difference in spelling in fchk files from g03 and g09:
     # "independ__e__nt" versus "independ__a__nt".
-    nbasis_indep = fchk.get("Number of independant functions",
-                            fchk.get("Number of independent functions", nbasis))
+    # The nbasis_indep is not used, so it is commented because of pylint unused-variable error
+    # nbasis_indep = fchk.get("Number of independant functions",
+    #                         fchk.get("Number of independent functions", nbasis))
```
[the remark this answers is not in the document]
rayhe88: It's done, a remove the lines

iodata/formats/fchk.py
```diff
@@ -435,3 +439,225 @@ def _triangle_to_dense(triangle: np.ndarray) -> np.ndarray:
...
+
+
+def _dump_real_arrays(name: str, val: np.ndarray, f: TextIO):
+    """Dumper for a array of float."""
+    nval = val.size
+    if nval != 0:
+        np.reshape(val, nval)
+        print("{0:40}   R   N={1:12}".format(name, nval), file=f)
+        k = 0
+
```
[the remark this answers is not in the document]
rayhe88: Empty line removed

iodata/formats/fchk.py
rayhe88: Add some lines for check if the number of occupations for orbitals alpha and beta are all integers. We need to work more in this line for the IOData object

iodata/formats/fchk.py
rayhe88: the changes are made

iodata/formats/fchk.py
rayhe88: I removed all related to `dump_many`

iodata/test/test_fchk.py
FarnazH: These two tests, and the other two which are all now marked with `@pytest.mark.xfail`, were commented out because they were converting WFN/WFX file formats containing post-HF wave-functions (and fractional orbital occupations and natural orbitals) to FCHK. So, when dumping FCHK and then loading it to compare to the initial WFN/WFX, the MO occupations do not match, so I commented these tests and opened issue #174 (not knowing that FCHK cannot store natural orbitals). I am not sure how this needs to be fixed, and in the meanwhile we can probably just raise an error for now, until we establish our conventions.

iodata/formats/fchk.py
FarnazH: @tovrstra, to add the second check (occupied followed by virtual orbitals), I add a check to make sure `mo.kind` is not "generalized". Does it fully cover this check? 

I am in favor of factoring out IOData preparing before passing it to `dump_one`, and good to have an issue on that.

PR #165, opened 2020-07-26
Pure python overlap.py

BradenDKelly:
This is part 1 of 2. 

Purpose: make pure python as close to the speed of cython as possible.

In this code shells, then primitives, then angular momentums are looped over. This code uses np.frompyfunc to define a vectorized single dimensional overlap calculation function. For two primitives interacting all angular momentums and x,y,z are then broadcast to this function. rather than 8 for loops, this uses 6.

In Part 2 - currently in progress, the loops over primitives will also be included in the broadcasting. Rather than 8 for loops this uses 4.

**Test 1: loading big.molden using (old) i7 2.4 GHz desktop**
python + cython: 33.29 seconds,  average over 50 iterations
**Pure python timings:**
No shell or primitive screening: 93.49 seconds,  average over 10 iterations
With shell screening:  52.14 seconds,  average over 10 iterations
With shell and primitive screening: 49.76 seconds,  average over 50 iterations

**Test 2: overlaps on small molecule - water using sto3g**
python + cython:  0.0116 seconds, average over 1000 iterations
pure python: 0.0299 seconds,  average over 1000 iterations

on a new laptop(windows) pure python is 29 seconds for the big.molden. I was not able to get cython to run on windows. (old laptop has a linux partition)


BradenDKelly:
@tovrstra If you could review this one, that would be great. I am not sure why, but I do not have the option of adding reviewers to any of my theochem/iodata PR's.

tovrstra:
OK, good to know. I've added myself as reviewer. I'll take a look soon.

iodata/overlap.py
```diff
@@ -81,16 +86,47 @@ def compute_overlap(obasis: MolecularBasis, atcoords: np.ndarray) -> np.ndarray:
...
-                for iexp1, (a1, cc1) in enumerate(zip(shell1.exponents, shell1.coeffs[:, 0])):
-                    s1 = scales[i1][iexp1]
-                    n0 = np.vstack(list(iter_cart_alphabet(shell0.angmoms[0])))
-                    n1 = np.vstack(list(iter_cart_alphabet(shell1.angmoms[0])))
-                    add_overlap(cc0 * cc1, a0, a1, s0, s1, r0, r1, n0, n1, result)
+            a0 = np.min(shell0.exponents)
+            a1 = np.min(shell1.exponents)
+            # arrays of angular momentums [[2, 0, 0], [0, 2, 0], ..., [0, 1, 1]]
+            n0 = np.array(iter_cart_alphabet2(shell0.angmoms[0]))
+            n1 = np.array(iter_cart_alphabet2(shell1.angmoms[0]))
```
tovrstra, line 94: Can these two go into the `if prefactor > 1.e-15:` block? I don't expect much performance gain. It just makes things easier to follow.

iodata/overlap.py
```diff
@@ -156,6 +214,17 @@ def gob_cart_normalization(alpha: np.ndarray, n: np.ndarray) -> np.ndarray:
         The normalization constant for the gaussian cartesian basis.
 
     """
-    vfac2 = np.vectorize(factorialk)
-    return np.sqrt((4 * alpha)**sum(n) * (2 * alpha / np.pi)**1.5
-                   / np.prod(vfac2(2 * n - 1, 2)))
+    vfac2 = np.vectorize(factorial2)
```
tovrstra, line 217: I'm not sure if this is needed. It may be a leftover from the old implementation. The `factorial2` function is already vectorized, e.g.

```python
>>> from scipy.special import factorial2
>>> factorial2([3, 5])
array([ 3., 15.])

```

iodata/overlap.py
```diff
@@ -156,6 +214,17 @@ def gob_cart_normalization(alpha: np.ndarray, n: np.ndarray) -> np.ndarray:
...
+    return np.sqrt((4 * alpha) ** sum(n) * (2 * alpha / np.pi) ** 1.5
+                   / np.prod(vfac2(2 * n - 1)))
+
+
+def iter_cart_alphabet2(n):
```
tovrstra, line 222: Would it be possible to reuse the function in `iodata/obasis.py`? See https://github.com/theochem/iodata/blob/master/iodata/basis.py#L318

tovrstra:
Do you expect `np.frompyfunc` to give performance improvements. As far as I understand it only introduces some convenience from NumPy ufuncs, but not necessarily performance. How does this compare to the double loop?

iodata/overlap.py
BradenDKelly: For sure, I should have caught that.

iodata/overlap.py
BradenDKelly: I will look more into it, but when I tried it did not work since that produces an iterable, and I needed to be able to pass explicit arrays to the vectorized function. I may have simply implemented it wrong.

BradenDKelly:
Yes, `np.frompyfunc` gives nice improvements, `np.vectorize` does not. My crude understanding is that `np.frompyfunc` is the stripped down spartan version of  `np.vectorize`. `np.vectorize` is the same speed as double for loops. Currently `np.frompyfunc` speeds up the code by 15-20%. My plan is to further pass it the for loops at the primitive level. I have simple test code which works, and this additional vectorization is twice as fast as the current vectorized approach. Implementing this is mostly done, but has a small bug making 5% of the overlaps incorrect, so I have to finish this part and then I can time the actual code. I can give you access to this branch if you would like? you could probably spot the bug... I seem to keep looking it over...

I tried vectorizing the `def compute_overlap_gaussian_1d`  function by padding the arrays and doing matrix-vector algebra to get rid of those for loops and it was disastrous. The creation of temporary arrays defeated the purpose. Numpy vectorization only works well on arrays of length hundreds or thousands. These ones over `ang_mom` are less than 10.

iodata/overlap.py
BradenDKelly: I can change that and time it to make sure, but, I believe you are correct. I tried replacing `np.vectorize` with `np.frompyfunc` since the latter makes things faster and the former makes them easier but is no faster than loops. It did not change the speed - therefore, I believe you are correct and neither was doing any additional vectorization.

iodata/overlap.py
```diff
@@ -67,8 +64,16 @@ def compute_overlap(obasis: MolecularBasis, atcoords: np.ndarray) -> np.ndarray:
...
 
+    n_max = np.max([np.max(shell.angmoms) for shell in obasis.shells])
+    go = GaussianOverlap(n_max)
+
+    # define a python ufunc (numpy function) for broadcasted calling over angular momentums
+    compute_overlap_1d = np.frompyfunc(go.compute_overlap_gaussian_1d, 5, 1)
+
     # Loop over shell0
     begin0 = 0
+    count = 0
```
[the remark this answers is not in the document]
BradenDKelly: It should be gone. It used to have a use and I unwittingly removed its use without noticing it. Good catch.

tovrstra:
Thanks for explaining the vectorization. This makes sense. Let's merge this one first before diving into the next improvement.

iodata/overlap.py
tovrstra: ok, makes sense.

iodata/overlap.py
tovrstra: Would the following be ok?

```python
>>> import numpy as np
>>> from iodata.basis import iter_cart_alphabet
>>> np.array(list(iter_cart_alphabet(3)))
array([[3, 0, 0],
       [2, 1, 0],
       [2, 0, 1],
       [1, 2, 0],
       [1, 1, 1],
       [1, 0, 2],
       [0, 3, 0],
       [0, 2, 1],
       [0, 1, 2],
       [0, 0, 3]])
```

BradenDKelly: Yes, that is a much more pythonic way. I have changed it.

iodata/overlap.py
```diff
@@ -18,17 +18,14 @@
...
+from scipy.special import binom, factorial2
 
-from .overlap_accel import add_overlap
 from .overlap_cartpure import tfs
 from .basis import convert_conventions, iter_cart_alphabet, MolecularBasis
 from .basis import HORTON2_CONVENTIONS as OVERLAP_CONVENTIONS
 
-
-__all__ = ['OVERLAP_CONVENTIONS', 'compute_overlap', 'gob_cart_normalization']
+__all__ = ['OVERLAP_CONVENTIONS', 'compute_overlap_py', 'gob_cart_normalization']
```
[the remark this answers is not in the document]
BradenDKelly: Changed as requested.

BradenDKelly:
I have made all of the changes requested. I made one additional cosmetic change on line 122 of `overlap.py`

I changed 

                        v = prefactor * np.prod(compute_overlap_1d(rn_0,
                                                                   rn_1,
                                                                   n0[:, None, :],
                                                                   n1[None, :, :],
                                                                    two_at),
                                                axis=2
                                                )

to:

                        v = np.prod(compute_overlap_1d(rn_0, rn_1, n0[:, None, :], n1[None, :, :],
                                                       two_at), axis=2)
                        v *= prefactor

I can easily reverse this if you prefer.

BradenDKelly:
@tovrstra please see my email before merging.

tovrstra:
@BradenDKelly Thanks. It works now. I managed to push a few commits.

tovrstra:
Note: CI fails because the conda servers are overloaded. This seems to happen every now and then. See e.g. https://github.com/conda/conda/issues/7800 This should be a transient issue.

tovrstra (approved):
Thanks for all the nice work. I have no further comments and our CI script passed on my local install (skipping the conda update, which is broken by an anaconda server overload). @FarnazH @BradenDKelly unless you have further thoughts about this one, it can be merged.

FarnazH (dismissed):

PR #167, opened 2020-07-28
Add contents table for supported file type

tczorro:
Enhancement: Add `gen_formats_tab.py` to generate a rst-format
table file `format_tab.inc`. This file is then included in the
new added page `supported_format.rst`
1. Add format and properties supported table
2. Add a temperary page for exhibiting the table
3. Add the autogen script to gen_docs.sh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ iodata/overlap_accel.c
 doc/_build
 doc/pyapi
 doc/formats.rst
+doc/formats_tab.inc
 
 # Distribution / packaging
 .Python
--- a/doc/_static/css/override.css
+++ b/doc/_static/css/override.css
@@ -39,5 +39,5 @@ table.docutils tbody tr.row-odd td {
 }
 
 div.wy-table-responsive {
-    max-height: 600px;
+    max-height: 90vh;
 }
--- a/doc/_static/css/override.css
+++ b/doc/_static/css/override.css
@@ -1,0 +1,48 @@
+/** css/override.css **/
+
+
+/* This line is theme specific - it includes the base theme CSS */
+@import 'theme.css';       /* for the Read the Docs theme */
+
+table.docutils {
+    border-collapse: separate !important;
+}
+
+th.head {
+    position: sticky;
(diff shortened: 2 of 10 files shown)


FarnazH:
@tovrstra we have a file format table thanks to Derrick; please let us know what you think. The order of columns might not be idea, but they are ordered based on the number of properties they load... The rows are also ordered based on the number formats that use them. So, the upper left part of the table is the most populated. The columns are cross-references, but the rows are not yet.

The main caveat is that derived attributes like `natoms` and `nelec` (that are available for almost all formats) are not among the `guaranteed` & `ifpresent` attributes of formats, so they are not checked in the table. I am not sure what is the best way to fix this. Any thoughts?

tovrstra:
Sorry, I lost track of this one. I'll add myself as reviewer.

FarnazH:
IOData property attributes (e.g. natom, nelec, charge, spinpol, atcorenums), which are derived from other attributes, are marked as always present in the format table, except for `fcidump` and `gaussianlog` formats (because they cannot be derived from the current loaded information). The only exception is `natom` & `spinpol` attributes which are available from `fcidump`.

FarnazH:
@tczorro and @tovrstra this is ready to be reviewed and merged. We might decide to put rows in alphabetical order (based on the IOData attribute names), so that it is easier to look through. The only remaining task is cross-referencing row attributes, but that needs some changes to IOData docstring format, so I believe it can be done later.

FarnazH:
@tovrstra I like the changes as it makes the table and make it more informative. I am going to rebase it now.

FarnazH:
I just added a `:` between file format and LD to make it a bit more clear. I tried to reduce the width of the header, but couldn't make it work in a short time.

tovrstra (approved):

tovrstra (approved):

tovrstra:
@FarnazH There may be a way to fix this, hang on ...

tovrstra:
@FarnazH Now the table is as high as it reasonably can on one's screen. Merging...

PR #170, opened 2020-07-28
Support pure H functions in molden file

### LINKED ISSUE #155: Fix loading Molden files with spherical high angular momentum basis functions
At the moment, it seems we don't load Molden files from PSI4 properly when they contain spherical basis functions with angular momentum above g. For such basis functions, the Molden format is not explicitly defined, so it remains a bit ambiguous how to handle them correctly.

Example: [example.zip](https://github.com/theochem/iodata/files/4875369/example.zip) (This was reported to be a problem with HORTON 2, but I assume the same issue exists in IOData, needs to be checked though. A smaller example would be needed for the unit tests.) 

I guess the following extension to the Molden format is needed to load this example correctly. We should interperpret the `[9G]` line as: "functions with angular moment g and higher are spherical", instead of only letting it affect the g functions.

tovrstra:
Fixes #155. I've included smaller test files which had the same problem.


tovrstra:
@evohringer Do you use ORCA often? This PR has an uncertainty related to ORCA Molden files. It would be easy to fix it with a Molden file from ORCA for a HF/cc-pVQZ calculation on a neutral Zn atom. Would you or someone in your group be able to generate one easily?

evohringer:
@tovrstra I just emailed you the molden file from orca.

tovrstra:
Few notes for the review:
- I've remove the comparison of conventions in `compare_mols` because it breaks load_dump_consistency tests when combining files from different formats. These asserts were not very useful in the first place.
- I've used `pytest.mark.parametrize` to remove some redundancies in `test_molden.py`

tovrstra:
I made just some minor improvements to the comments and I'll merge after passing tests.

Issue #36, opened 2019-01-25
Steal all load functions from tamkin


tovrstra:
See
- https://github.com/molmod/tamkin/tree/master/tamkin/io
- https://github.com/molmod/tamkin/blob/master/tamkin/test/test_io.py
- https://github.com/molmod/tamkin/tree/master/tamkin/data/test

fwmeng88:
I am working this issue by starting with CHARMM file format, https://github.com/molmod/tamkin/blob/a15141745c90abbe8dadb1394f538e30a3523d48/tamkin/io/charmm.py#L51-L138. I am wondering if you can help with some questions regarding CHARMM file format. 
1. Should the load functions be dependent or independent of tamkin? I think tamkin is built with Python 2 where we run over some exceptions errors when testing. For example, the `print` in https://github.com/molmod/tamkin/blob/a15141745c90abbe8dadb1394f538e30a3523d48/tamkin/partf.py#L1019 will result in an error in Python 3.7.
2. I am not sure which kind of information should be returned at the end of functions. I just copied the `load_molecule_charmm` and `load_coordinates_charmm` where I don't know if there is any need for `load_peptide_info_charmm`. I didn't take the `load_modes_charmm` bacuase I think the normal mode analysis is not our intention.
I will try to port codes from _cpmd.py, gamess.py, gaussian.py, qchem.py and vasp.py_ in the coming days.
Thank you very much! @tovrstra @FarnazH

PaulWAyers:
It's my impression that the goal is to steal the wavefunction-loading capabilities of tamkin.  Tamkin should not be a dependence (just rewrite clean Python 3 code using tamkin code as a starting point); and I would not have started with CHARMM, which I think is probably not the point at all of this issue, since HORTON is a wavefunction method and I am do not think CHARMM has wavefunction information.  Certainly load_peptide_info_charmm is not needed (or even useful, I suspect).

@tovrstra knows better what CHARMM has, if anything, that is useful in the iodata context. Perhaps only the geometry and the charge?

FarnazH:
I think it would be good to start with something simpler like `qchem.py`.

As @PaulWAyers mentioned, the code you port shouldn't add any extra dependency. I also think any information in the output files are worth extracting, as `iodata` is a parsing module in some sense.  I'll leave specific questionss regarding `CHARMM` to @tovrstra.

fwmeng88:
[...]

PR #171, opened 2020-07-28
Add load_one function for CHARMm crd files.

lmacaya:
This PR is associated with #36  to support CHARmm crd files. Particularly, segment id, residue id and weight columns (used to store a mass or charge) are stored in extra dict. Perhaps for the latter, it is convenient to apply a conditional and store them in atmassess or atcharges as appropriate.


iodata/formats/charmm.py
```diff
@@ -0,0 +1,138 @@
...
+                'weights': weights
+            }
+            result = {
+                'atcoords': atcoords,
+                'atffparams': atffparams,
+                'extra': extra,
+                'title': title,
+            }
+            break
+    if title_end is False:
```
tovrstra, line 101: minor comment:
```suggestion
    if not title_end:
```

iodata/formats/charmm.py
```diff
@@ -0,0 +1,138 @@
...
+            atffparams = {
+                'attypes': attypes,
+                'atnumbers': atnumbers,
+                'resnames': resnames,
+                'resnums': resnums
+            }
+            extra = {
+                'segid': segid,
+                'resid': resid,
+                'weights': weights
```
tovrstra, line 92: Could you use the `IOData.atmasses` attribute instead? Also, conversion to atomic units would be needed. I'll show it below. Also, weight is technically speaking a force as opposed to a mass. Just to avoid unit confusion, it is better not to mix weight and mass terminology.

iodata/formats/charmm.py
```diff
@@ -0,0 +1,138 @@
...
+        atnumbers.append(int(words[0]))
+        resnums.append(int(words[1]))
+        resnames.append(words[2])
+        attypes.append(words[3])
+        pos[i, 0] = float(words[4])
+        pos[i, 1] = float(words[5])
+        pos[i, 2] = float(words[6])
+        segid.append(words[7])
+        resid.append(int(words[8]))
+        weights.append(float(words[9]))
```
tovrstra, line 136:
```suggestion
        atmasses.append(float(words[9]) * amu)
```
The amu constant can be imported `from iodata.utils`. In the unit test, the same conversion is then needed, e.g.

```python
assert_allclose(mol.atmasses[-1], 15.99900 * amu)
```

or even 

```python
assert mol.atmasses[-1] == 15.99900 * amu
```

because they should match exactly.

iodata/formats/charmm.py
```diff
@@ -0,0 +1,138 @@
...
+            resnums = np.array(data[1])
+            resnames = np.array(data[2])
+            attypes = np.array(data[3])
+            atcoords = data[4]
+            segid = np.array(data[5])
+            resid = np.array(data[6])
+            weights = np.array(data[7])
+            atffparams = {
+                'attypes': attypes,
+                'atnumbers': atnumbers,
```
tovrstra, line 85: `atnumbers` would be easily confused with `IOData.atnums`. Would `atserial` or `atindex` also be ok? (If these are always just range(1, natom+1), then it would not be necessary to load them, but I'm not sure if it can be something else in practice.)

iodata/test/test_charmm.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_load_crambin():
+    # test gro file of one water
+    with path('iodata.test.data', 'crambin.crd') as fn_crd:
+        mol = load_one(str(fn_crd))
+    check_crambin(mol)
+
+
+def check_crambin(mol):
```
tovrstra, line 39: There is no need for a separate check function, because it is not reused by other tests. Could you also add more assert lines, e.g. for a few elements of `atcoords` and the shapes of the values of the `atffparams` and `extra` attributes?

iodata/formats/charmm.py
lmacaya: I agree. I only used this name based on the documentation. I will change it.

iodata/formats/charmm.py
lmacaya: That is correct. It is just a range for the number of atoms. I wil skip it.

iodata/test/test_charmm.py
lmacaya: No problem! Working on that.

PR #172, opened 2020-07-28
Add load_one and load_many functions for Gromacs gro files.

lmacaya:
This PR is associated with #37 to support Gromacs gro files. I take most everything for molmod without many modifications, including some units that Gromacs use by default (nanometers and picoseconds).


iodata/formats/gromacs.py
```diff
@@ -0,0 +1,148 @@
...
+__all__ = []
+
+
+PATTERNS = ['*.gro']
+
+
+@document_load_one('GRO', ['atcoords', 'atffparams', 'cellvecs', 'extra', 'title'])
+def load_one(lit: LineIterator) -> dict:
+    """Do not edit this docstring. It will be overwritten."""
+    frame_found = False
```
tovrstra, line 44: Related to `_helper_read_frame return None` instead of exception: line can be removed.

iodata/test/test_gromacs.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def check_water(mol):
+    """Test some things on a water file."""
+    assert mol.title == 'MD of 2 waters'
+    assert mol.atcoords.shape == (6, 3)
+    assert mol.atffparams['attypes'][2] == 'HW3'
+    assert mol.atffparams['resnames'][-1] == 'WATER'
+    assert_equal(mol.atffparams['resnums'][2:4], [1, 2])
+    assert_allclose(mol.cellvecs[0][0], 1.82060 * nanometer, atol=1.e-5)
```
tovrstra, line 47: Could you also add a check for the value of an atomic position?

iodata/formats/gromacs.py
lmacaya: Good suggestion! Working on it.

iodata/test/test_gromacs.py
lmacaya: Also, I will add a test for velocity values.

iodata/formats/gromacs.py
```diff
@@ -0,0 +1,145 @@
...
+        time = float(line[pos + 2:]) * picosecond
+    else:
+        time = 0.0
+    # Read the second line for number of atoms.
+    natoms = next(lit)
+    if natoms is not None:
+        try:
+            natoms = int(natoms)
+        except ValueError:
+            return None
```
FarnazH, line 109: Can you add a comment in the code to explain why you need a `try/except` statement here? 
I don't have any experience with `GRO` files, but I am thinking that `if natoms is not None`, then you can convert it into `int`... And I am not sure why there is a `return None`. Does that specify that there is a `Bad file`?

iodata/formats/gromacs.py
```diff
@@ -0,0 +1,145 @@
...
+        except IOError:
+            return
+
+
+def _helper_read_frame(lit: LineIterator) -> Tuple:
+    """Read one frame."""
+    # Read the first line, get the title and try to get the time.
+    # Time field is optional.
+    line = next(lit)
+    title = line.split(',')[0] if 't=' in line else line[:-1]
```
FarnazH, line 97: If time is not given in the title, I am confused why line is reversed...
```suggestion
    title = line.split(',')[0] if 't=' in line else line
```

iodata/formats/gromacs.py
```diff
@@ -0,0 +1,145 @@
...
+    pos = line.rfind('t=')
+    if pos >= 0:
+        time = float(line[pos + 2:]) * picosecond
+    else:
+        time = 0.0
```
FarnazH, line 102: Possible simplification (what you had is also good):
```suggestion
    time = 0.0
    if 't=' in line:
        time = float(line.split('t=')[1]) * picosecond
```

iodata/test/test_gromacs.py
```diff
@@ -0,0 +1,61 @@
...
+# IODATA is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>
+# --
+# pylint: disable=unsubscriptable-object
+"""Test iodata.formats.orcalog module."""
```
FarnazH, line 20: Please update this:
```suggestion
"""Test iodata.formats.gromacs module."""
```

iodata/formats/gromacs.py
lmacaya: I will add it. Regarding the second question, that is correct. This is a simple test to evaluate that the number of atoms in the file is present in the second row. If this line is anything else, we can assume that the format is incorrect. Obviously, I can add more checkpoints, for me there is no problem at all.

iodata/formats/gromacs.py
lmacaya: It takes the complete line except the new line character. Just aesthetics. I can change it 👍

iodata/formats/gromacs.py
lmacaya: Nice! It looks great. Thanks for the suggestion.

iodata/test/test_gromacs.py
lmacaya: Oops. I forgot to take a second look at the template that I used. Sorry.

iodata/formats/gromacs.py
FarnazH: Sorry, for some reason I thought you are using `line[::-1]` which reverses the string, but you had only one `:`. Removing the new line character is good, and you shouldn't remove it.

iodata/formats/gromacs.py
FarnazH: These checks are good, but actually don't do checks like this because then there are so many checks that you need to include. For example, the first line of XYZ file is also supposed to be an integer (for number of atoms), but in loading an XYZ file we assume it is the case. See: https://github.com/theochem/iodata/blob/master/iodata/formats/xyz.py#L101
If an XYZ file is incorrect, and the first line isn't an integer, the `natom = int(next(lit))` would fail, and the error should be clear-ish on what went wrong.

I think it is safe to use `natom = int(next(lit))` here as well, then you can also simplify the `load_one` a bit, by removing try/except and if statements because data would not be `None`, see: https://github.com/theochem/iodata/pull/172/files#diff-e32c760d08263acbc390686b11208558R47-R51

lmacaya: Thanks for the meaningful and explanatory reply! I really appreciate it, because it allows me to understand a little more about the module's work philosophy. I will commit these changes as soon as possible.

lmacaya: > I think it is safe to use `natom = int(next(lit))` here as well, then you can also simplify the `load_one` a bit, by removing try/except and if statements because data would not be `None`, see: https://github.com/theochem/iodata/pull/172/files#diff-e32c760d08263acbc390686b11208558R47-R51

Finally, I removed the if statement only. I was using the try/except statement for the load_many function in case that the gro file was a trajectory.

FarnazH:
Thanks @lmacaya. I just noticed that `master` has been merged into your branch; see 
be0ac77 above... Can you please remove that? Then, we are good to merge. Thanks a lot.

Issue #146, opened 2020-06-23
obasis documentation/features


PaulWAyers:
Executive summary: improve documentation; allow label-to-basis conversion; convert-to-primitives functionality.

We should make sure that obasis (both code and docs) are thoroughly documented. (We've had several questions about it in the last few days).

It would be useful to have a label-to-basis-index and basis-index-to-label functionality, similar to PySCF. It would not need to be implemented exactly like this, but it would be nice to have functionality that, for a given basis function (say, the 17th) would tell you its type, contraction coefficients, exponents, center etc. (e.g., that it was the 3rd dz^2 basis function on the second atomic center). Similarly, it would be nice to be able to ask for a specific type of basis function (e.g., the 3rd dz^2 on a specified atom) and return the basis index. 

Finally, it would be nice to have a tool to convert a contracted basis set (and its MOs) into a completely decontracted (100% primitives) format. This is especially useful for dumping *.wfx and *.wfn. It's also useful (very occassionally) since decontracted basis sets are often preferable for small, highly-charged polycations (e.g., isoelectronic series).

FarnazH:
Would it be also useful to make the functionality for converting between Cartesian and Pure types more transparent and user-friendly, so that the user can convert between them, and choose what type to dump when writing out a file format?

PaulWAyers:
Yes. And it would be nice to have a lot of other functionality. E.g.,
1. Convert from L1 normalized to L2 normalized to un-normalized basis set conventions. It's especially important since integrals and *.wfn and *.wfx usually use unnormalized Gaussians.
2. [from Farnaz] Convert from pure-to-Cartesian basis sets.
3. [from my first issue] Convert from contracted to uncontracted basis sets. (Also useful for *.wfn and *.wfx, but also the cse where you want to take a neutral-atom basis and apply it to a highly charged system).

PaulWAyers:
@tovrstra do you have notes on the L1 norm for spherical Gaussian basis functions? The L1 normalization is "supported" but not really documented.

tovrstra:
It is not implemented yet. The field was introduced because it seemed useful for density-fitting basis sets, but it was not needed yet with the current file formats.

PaulWAyers:
[...]

PR #175, opened 2020-08-05
Molecular Basis

Ali-Tehrani:
This pull request is regarding issue (#146).

Added
---------
- Conversion to decontracted molecular basis (c97cad3)
- Ability to generally convert a vector between two basis sets, given that one knows the overlap matrices (6f3f76f)
- Method in `MolecularBasis` to convert all primitive shells in a contracted shell to a specific kind (either Cartesian or Pure) (345f190). 
    - Ran into the problem of iodata assuming all coefficients in a primitive shell are the same. However, conversion to a different            basis varys the coefficients across a primitive shell. This is fixed by averaging them.
- General function to convert a kind of a single primitive shell to Cartesian or pure (345f190) . 
- Update the code document with definitions, and equations (d956cfa, 8eb2ea7).

### DIFF
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -176,7 +176,7 @@ class Shell:
 
     @property
     def nprim(self) -> int:  # noqa: D401
-        """Number of primitive contracted shells, also known as the contraction length."""
+        """Number of primitive, also known as the contraction length."""
         return len(self.exponents)
 
     @property
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -38,11 +38,13 @@ ANGMOM_CHARS = 'spdfghiklmnoqrtuvwxyzabce'
 
 def _alsolist(f):
     """Wrap a function to accepts also list as first argument and then return list."""
+
     @wraps(f)
     def wrapper(firsts, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


PaulWAyers:
Exactly. It would be nice if we could read in the *.wfn and *.wfx and, to the extent possible, reconstruct the true computational basis. That would speed subsequent calculations appreciably. 

I agree that this is a lot of code to put in `IOData`. I'm not sure where else to put it, though; it doesn't really belong in any module, as the main utility of basis-set-changing is when post-processing one calculation as a pre-processing step for a subsequent calculation. It might be that it belongs in `HORTON3` proper, though I'm not sure how the workflow would be:  probably `IOData` to `gbasis` to `IOData`. Given that the feature of `gbasis` that is being used is already (redundantly, for the sake of modularity) in `IOData`, putting this basis-set-conversion functionality in `IOData` seems semi-sensible (but, I agree, not perfectly so).

tovrstra:
I'm closing this PR because several aspects need more discussion before the implementation and because too many different issues and features are being addressed in one PR, making it hard to fix all problems before merging. I've opened several new issues related to this PR, making references to the code here when relevant.

PR #187, opened 2020-08-19
Document Github, git and SSH configuration

tovrstra:
This is mainly intended to make the contributing guide more accessible. The instructions assumed an SSH key was uploaded to one's Github account, but this is not the case by default, which may cause confusion.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -240,7 +240,8 @@ to avoid duplicate efforts.
 
      .. code-block::
 
-       Host github.ugent.be
+       Host github.com
+           Hostname github.com
            ForwardX11 no
            IdentityFile /home/your_user_name/.ssh/id_rsa_github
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -203,23 +205,57 @@ formats, this could become more complicated.
...
+   - Upload your *public* SSH key to `<https://github.com/settings/keys>`_.
+     This is a single long line in ``id_rsa_github.pub``, which you can
+     copy-paste into the browser.
+
+   - Configure SSH to use this key pair for authentication when pushing
+     branches to Github. Add the following to your ``.ssh/config`` file:
+
+     .. code-block::
+
+       Host github.ugent.be
```
FarnazH, line 243: I think this should be "Host github.com", no?

tovrstra: oops

tovrstra: @FarnazH Thanks for catching this one! I've also add the Hostname line. It may be redundnant in shome SSH versions, but it makes things a little clearer.

PR #194, opened 2020-08-24
Update README & Add License Shield

FarnazH:
This PR:

1. Updates the dependencies in README
2. Add License shield and remove the License section of README (it makes it shorter & more informative)

I tried to update the python version shield, but I think that is tied to pypi (which based on its last release still supports python 2.7 & 3).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,10 +51,9 @@ The following dependencies will be necessary for IOData to build properly,
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


tovrstra:
Indeed, as soon as we make a stable release on pypi, the py version shield will be updated too.

tovrstra (approved):
LGTM!

PR #195, opened 2020-08-25
Handle warnings in unit tests

tovrstra:
Several unit tests were generating warnings that were uncatched. I have also removed one warning from `molekel.py`, because it was raised in a fairly harmless scenario. I've also changed the code a bit: the entire charge section is omitted when there are no charges to be written.

### DIFF
--- a/iodata/test/common.py
+++ b/iodata/test/common.py
@@ -18,17 +18,26 @@
 # --
 """Utilities for unit tests."""
 
-from os import path
+import os
 from contextlib import contextmanager
 
 import numpy as np
 from numpy.testing import assert_equal, assert_allclose
+import pytest
 
+from ..api import load_one
 from ..overlap import compute_overlap
 from ..basis import convert_conventions
+from ..utils import FileFormatWarning
+
+try:
+    from importlib_resources import path
+except ImportError:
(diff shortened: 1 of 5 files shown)


tovrstra:
@evohringer could you tell me which one it is? It seems they are all consistent. If you go to "Files changed" and hover over a green line, a blue plus will appear, which can be used to make local comments.

iodata/test/common.py
```diff
@@ -158,3 +167,32 @@ def check_orthonormal(mo_coeffs, ao_overlap, atol=1e-5):
     message = 'Molecular orbitals are not orthonormal!'
     assert_allclose(mo_overlap, np.eye(mo_count),
                     rtol=0., atol=atol, err_msg=message)
+
+
+def load_one_warning(filename: str, fmt: str = None, match: str = "", **kwargs):
```
FarnazH, line 172: A very minor (and probably picky) comment: I would use `None` as the default value of `match` instead of `""` which is mutable. I checked `pytest.warns` documentation, and that also uses `None` as default.

tovrstra: Thanks for catching this. It's fixed.

PR #188, opened 2020-08-20
Input Writer

FarnazH:
This PR addresses issue #43. 
1. The`write_input` is added to `api` module.
2. The Gaussian input writer is added as an example and is tested with the default and user-defined input template.

PR #196, opened 2020-08-25
Fix "unsupported operand" issue when accessing IOData.charge

tovrstra:
This is the fix for the following type of error message:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/maarten/anaconda3/envs/iodata/lib/python3.8/site-packages/iodata/iodata.py", line 274, in charge
    return self.atcorenums.sum() - self.nelec
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType
```

This issue is hampering two PRs (#188 and #186), so I'm putting in a separate PR, which can be merged easily. 

(Related issues with nelec and spinpol are also fixed.)

### DIFF
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -114,7 +114,8 @@ class IOData:
         e.g. for a wire. Two vectors describe a 2D cell, e.g. for a membrane.
         Three vectors describe a 3D cell, e.g. a crystalline solid.
     charge
-        The net charge of the system.
+        The net charge of the system. When possible, this is derived from
+        atcorenums and nelec.
     core_energy
         The Hartree-Fock energy due to the core orbitals
     cube
@@ -266,18 +267,24 @@ class IOData:
     def atcorenums(self, atcorenums):
         self._atcorenums = atcorenums
         if self._charge is not None:
-            self.nelec = self._atcorenums.sum() - self._charge
+            # _charge is treated as the dependent one, while atcorenums and
+            # nelec are treated as independent.
+            if self.nelec is None:
+                # Switch to storing _nelec.
+                self._nelec = atcorenums.sum() - self._charge
             self._charge = None
 
     @property
     def charge(self) -> float:
         """Return the net charge of the system."""
-        if self.atcorenums is not None and self.nelec is not None:
-            return self.atcorenums.sum() - self.nelec
-        return self._charge
+        # The internal _charge is used only if it cannot be derived.
+        if self.atcorenums is None or self.nelec is None:
+            return self._charge
+        return self.atcorenums.sum() - self.nelec
 
     @charge.setter
     def charge(self, charge: float):
+        # The internal _charge is used only if atcorenums is None.
         if self.atcorenums is None:
             self._charge = charge
         else:
@@ -289,8 +296,8 @@ class IOData:
         natom = None
         if self.atcoords is not None:
             natom = len(self.atcoords)
-        elif self.atcorenums is not None:
-            natom = len(self.atcorenums)
+        elif self._atcorenums is not None:
+            natom = len(self._atcorenums)
         elif self.atgradient is not None:
             natom = len(self.atgradient)
         elif self.atfrozen is not None:
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -248,6 +248,15 @@ class IOData:
     two_ints: dict = {}
     two_rdms: dict = {}
 
+    def __attrs_post_init__(self):
+        if self._charge is not None:
+            # Trigger charge setter to acchieve consistency in properties
+            # nelec, charge and atcorenums. This is needed because the
+            # attr constructor bypasses the setters. See
+            # https://www.attrs.org/en/stable/init.html#private-attributes
+            self.charge = self._charge
+
+
     # Public interfaces to private attributes
 
     @property
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -311,6 +311,8 @@ class IOData:
     @property
     def nelec(self) -> float:
         """Return the number of electrons."""
+        # When the molecular orbitals are present, they determine the number
+        # of electrons. Only when mo is absent, we use a stored value.
         if self.mo is not None:
             return self.mo.nelec
         return self._nelec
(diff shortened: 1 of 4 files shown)


iodata/iodata.py
```diff
@@ -259,22 +260,31 @@ def atcorenums(self) -> np.ndarray:
...
     @atcorenums.setter
     def atcorenums(self, atcorenums):
         self._atcorenums = atcorenums
+        if self._charge is not None:
+            # _charge is treated as the dependent one, while atcorenums and
+            # nelec are treated as independent.
+            if self.nelec is None:
+                # Switch to storing _nelec.
+                self._nelec = atcorenums.sum() - self._charge
+            self._charge = None
```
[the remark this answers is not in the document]
lmacaya: if I understood the code correctly, by assigning the variables `_atcorenums` and `_nelec`, it is possible to get the net charge value and even update it. If the charge was assigned before setting _atcorenums, it remains in _charges. By using `self._charge = None`, we get rid of the initial charge.

iodata/iodata.py
```diff
@@ -259,22 +260,31 @@ def atcorenums(self) -> np.ndarray:
...
+            self.atcorenums = result
         return result
 
     @atcorenums.setter
     def atcorenums(self, atcorenums):
         self._atcorenums = atcorenums
+        if self._charge is not None:
+            # _charge is treated as the dependent one, while atcorenums and
+            # nelec are treated as independent.
+            if self.nelec is None:
```
lmacaya, line 272:
```suggestion
            if self.nelec is None and atcorenums is not None:
```
Maybe this fix `AttributeError: 'NoneType' object has no attribute 'sum'`

iodata/iodata.py
```diff
@@ -259,22 +260,31 @@ def atcorenums(self) -> np.ndarray:
...
+        if self.atcorenums is None or self.nelec is None:
             return self._charge
         return self.atcorenums.sum() - self.nelec
 
     @charge.setter
     def charge(self, charge: float):
+        # The internal _charge is used only if atcorenums is None.
         if self.atcorenums is None:
             self._charge = charge
         else:
```
lmacaya, line 290: If data.charge is set to None, after setting data.atcorenums and data.charge, raise an error (Line 291):
```python
from iodata import IOData
d = IOData()
d.atnums = np.array([6, 1, 1, 1, 1])
d.charge = 1
d.charge = None
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lmacaya/Desktop/horton3/fix_charge_issue/iodata/iodata/iodata.py", line 291, in charge
    self.nelec = self.atcorenums.sum() - charge
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```
I don't know if this is an issue, but I point it out just in case.

iodata/iodata.py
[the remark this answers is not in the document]
tovrstra: I think I intended it the way Luis explained. It is just to clarify (thus not strictly necessary) that the `charge` is going to be derived from `nelec` and `atcorenums`.

iodata/iodata.py
```diff
@@ -259,22 +260,31 @@ def atcorenums(self) -> np.ndarray:
             # https://github.com/PyCQA/pylint/issues/1694
             # pylint: disable=no-member
             result = self.atnums.astype(float)
-            self._atcorenums = result
+            self.atcorenums = result
```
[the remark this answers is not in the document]
tovrstra: (Comments are only possible on lines that have changed, which is a bit unfortunate, because sometimes new changes imply things about old code.)

Agreed, `result` can be eliminated, making things shorter. I'm not sure why it was this way, probably related to some old code.

iodata/iodata.py
tovrstra: I'll check

iodata/iodata.py
tovrstra: Needs to be fixed. I guess this is comparable to one of @FarnazH 's comments. Thanks for bringing this up.

iodata/iodata.py
tovrstra: To get intuitive behavior, the code had to be extended, which is now the case in the latest commit. I've added several variations of unit tests to make sure it works (and checked with coverage that they reached all code paths.)

iodata/iodata.py
tovrstra: This is also fixed now.

iodata/iodata.py
```diff
@@ -252,31 +253,45 @@ class IOData:
...
+            self.atcorenums = self.atnums.astype(float)
+        return self._atcorenums
 
     @atcorenums.setter
     def atcorenums(self, atcorenums):
+        if atcorenums is None:
+            if self.nelec is not None:
+                # Set _charge because charge can no longer be derived from
+                # atcorenums and nelec.
+                self._charge = self._atcorenums.sum() - self.nelec
```
lmacaya, line 270: I think this line will raise an error if `nelec` was set before `atcorenums` and the latter is set to None afterwards.

tovrstra: Thanks. I'll add a unit test for this scenario, and make sure it works.

tovrstra: You were right! It should be fixed now. I also found another minor issue with the dtype of atcorenums, which is fixed too.

iodata/iodata.py
```diff
@@ -301,9 +318,11 @@ def natom(self) -> int:
     @property
     def nelec(self) -> float:
         """Return the number of electrons."""
-        if self.mo is None:
-            return self._nelec
-        return self.mo.nelec
+        # When the molecular orbitals are present, they determine the number
```
mcoolsce, line 321: If a IOData object is created with only `charge `and `atnums `present
```python
mol = IOData(charge=-1, atnums=[8, 1, 1])
print(mol.nelec)
```
printing out `nelec` results in `None` while, after accessing `atcorenums` first, it prints out the correct value of `11.0`.
```python
mol = IOData(charge=-1, atnums=[8, 1, 1])
print(mol.atnumcores)
print(mol.nelec)
```
Is this the correct behaviour, or should `nelec` also be derived from `atnums` and the `charge` only?

mcoolsce (commented):
Everything looks good to me, accept one possible inconsistency I could find.

iodata/iodata.py
tovrstra: Good point, and there was a way to fix it. pfew.

tovrstra:
@FarnazH @lmacaya @mcoolsce Thanks for reviewing everyone! It really helped a lot!

PR #197, opened 2020-08-25
Add GAMESS format

RichRick1:
Add new GAMESS format along with tests. Both script and test are transferred from MOLMOD.


tovrstra:
@RichRick1 Thanks for your contribution! The CI stumbled upon some linting issues, which should be easy to fix. See https://travis-ci.com/github/theochem/iodata/jobs/377467685#L495 and further below. If you need help with this, please let us know. If you like, you can run the CI locally to speed things up. See https://github.com/theochem/iodata/blob/master/CONTRIBUTING.rst#github-work-flow (Steps 2 and 9 reproduce most of the CI. This is may take some time upon the first run, because Roberto installs miniconda if not present yet in the default location.)

iodata/formats/gamess.py
```diff
@@ -0,0 +1,144 @@
...
+from ..utils import angstrom, LineIterator
+
+
+__all__ = []
+
+
+PATTERNS = ['*.gamessout']
+
+
+def _read_data(gms) -> tuple:
```
tovrstra, line 34: Very minor:
```suggestion
def _read_data(gms: LineIterator) -> tuple:
```
(similarly below.)

iodata/formats/gamess.py
```diff
@@ -0,0 +1,144 @@
...
+    N = len(res["symbols"])
+    # if the data are already read before, just overwrite them
+    gradient = res.get("gradient")
+    if gradient is None:
+        gradient = np.zeros((N, 3), float)
+    for i in range(N):
+        words = next(gms).split()
+        gradient[i, 0] = float(words[2])
+        gradient[i, 1] = float(words[3])
+        gradient[i, 2] = float(words[4])
```
tovrstra, line 84: The following appears to be working. (I've only recently learned this. We don't do it everywhere yet.)
```suggestion
        gradient[i] = words[2:5]
```

e.g.

```python
>>> import numpy as np
>>> a = np.zeros(3)
>>> a[:] = "1", "2", "3"
>>> a
array([1., 2., 3.])
```

tovrstra (commented):
Looks good, just a few very minor comments.

In general, I'd suggest to use variable names consistent with the rest of the code: `gms` -> `lit` and `res` -> `results`.

iodata/formats/gamess.py
```diff
@@ -0,0 +1,144 @@
...
+
+    coordinates = res.get("atcoords")
+    if coordinates is None:
+        coordinates = np.zeros((N, 3), float)
+    for i in range(N):
+        words = next(gms).split()
+        numbers[i] = int(float(words[1]))
+        coordinates[i, 0] = float(words[2]) * angstrom
+        coordinates[i, 1] = float(words[3]) * angstrom
+        coordinates[i, 2] = float(words[4]) * angstrom
```
FarnazH, line 68: The same feature @tovrstra mentioned below would be applicable here as well.
```suggestion
        coordinates = words[2:5] * angstrom
```

iodata/test/test_gamess.py
```diff
@@ -0,0 +1,62 @@
...
+    assert_equal(data.atnums[-1], 9)
+    assert_equal(data.atcoords.shape, (N, 3))
+    assert_allclose(data.atcoords[0, 1] / angstrom, -0.1843157808)
+    assert_allclose(data.atcoords[3, -1] / angstrom, 1.2926708150)
+    assert_allclose(data.atcoords[-1, 0] / angstrom, 3.8608437748)
+    assert_allclose(data.energy, -959.9675629527)
+    assert_equal(data.atgradient.shape, (N, 3))
+    assert data.atgradient[0, 1] - 1.5314677838E-05 < 1e-10
+    assert abs(data.atgradient[3, -1] - 8.5221217336E-06) < 1e-10
+    assert abs(data.atgradient[-1, 0] - 2.1211421041E-05) < 1e-10
```
FarnazH, line 52: Can you please add a few lines of tests for gradient points that have larger values?

iodata/test/test_gamess.py
```diff
@@ -0,0 +1,62 @@
...
+    from importlib.resources import path
+
+
+def test_load_one_gamess_punch():
+    with path('iodata.test.data', 'PCGamess_PUNCH.gamessout') as f:
+        data = load_one(str(f))
+    N = len(["CL", "H", "H", "H", "H", "F", "F", "F", "F", "H", "F"])
+    assert data.title == "Simple example sample optimization with Hessian output for Toon"
+    assert data.g_rot == "C1"
+    # # assert data["symbols"] == ["CL", "H", "H", "H", "H", "F", "F", "F", "F", "H", "F"]
```
FarnazH, line 39: Please remove this line.

iodata/formats/gamess.py
```diff
@@ -0,0 +1,144 @@
...
+        if line == " $END\n":
+            break
+        line = line[5:-1]
+        for j in range(len(line) // 15):
+            tmp[counter] = float(line[j * 15:(j + 1) * 15])
+            counter += 1
+    return hessian
+
+
+def _read_masses(gms, res):
```
FarnazH, line 107:
```suggestion
def _read_masses(gms, res) -> np.ndarray:
```

iodata/formats/gamess.py
```diff
@@ -0,0 +1,144 @@
...
+    counter = 0
+    while counter < N:
+        words = next(gms).split()
+        for word in words:
+            masses[counter] = float(word)
+            counter += 1
+    return masses
+
+
+@document_load_one("PUNCH", ['title', 'energy', 'grot', 'atgradient', 'athessian', 'atmasses'])
```
FarnazH, line 120:
```suggestion
@document_load_one("PUNCH", ['title', 'energy', 'grot', 'atgradient', 'athessian', 'atmasses', 'atnums', 'atcoords'])
```

iodata/test/test_gamess.py
RichRick1: It seems like the largest value of gradient in the test file is 3.3749561551e-05. Should I add this one to the test?

FarnazH:
@RichRick1, thanks. I have no further comments. @tovrstra I made the variable name changes you suggested.

iodata/formats/gamess.py
```diff
@@ -0,0 +1,141 @@
...
+
+PATTERNS = ['*.gamessout']
+
+
+def _read_data(lit: LineIterator) -> tuple:
+    """Extract ``title``, ``symmetry`` and ``symbols`` from the punch file."""
+    title = next(lit).strip()
+    symmetry = next(lit).split()[0]
+    if symmetry != "C1":
+        raise NotImplementedError("Only C1 symmetry is supported.")
```
tovrstra, line 39: I'm not sure if we really only support C1 symmetry here. Do we know someone who can generate a few GAMESS punch files to check this?

iodata/formats/gamess.py
```diff
@@ -0,0 +1,141 @@
...
+        gradient = np.zeros((natom, 3), float)
+    for i in range(natom):
+        words = next(lit).split()
+        gradient[i] = words[2:5]
+    return energy, gradient
+
+
+def _read_hessian(lit: LineIterator, result: dict) -> np.ndarray:
+    """Extract ``hessian`` from the punch file."""
+    assert "hessian" not in result
```
tovrstra, line 86: This may have to be `athessian`. How likely will this assert fail? If it is a consistency check for our code, an `assert` is fine, which I believe was the purpose here. If this is to detect broken punch files, `lit.error(...)` would be more suited, probably not the case here.

iodata/formats/gamess.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+
+from ..docstrings import document_load_one
+from ..utils import angstrom, LineIterator
+
+
+__all__ = []
+
+
+PATTERNS = ['*.gamessout']
```
tovrstra, line 31: This may not be the common extension. I believe it is usually `.dat`, but I'm not sure.

tovrstra (commented):
I have a few more questions. We may have to get in touch with a GAMESS user to resolve some of the issues.

iodata/formats/gamess.py
RichRick1: I googled this and it seems like .dat is the most popular output file. However, there are also .out and .log files too (according to this https://oban.spst.tju.edu.cn/education/qmcourse/v2010/week4/theory.php)

iodata/formats/gamess.py
PaulWAyers: GAMESS supports a lot of symmetries (only Abelian I think though) but I'm not sure what we support. We may just allow a failure if it isn't C1? In HORTON more generally we don't support symmetries in any real sense, certainly not in `grid` or `gbasis` at this point. (I guess we can "copy over" symmetries.) My vague recollection (and a very very very tiny bit of googling seems to confirm this) is that this file contains only symmetry-unique atoms, so we would be incapable of supporting non-C1 symmetry without significant additional coding. So I feel like the "only C1 symmetry is supported" is reasonable for now. We could probably add a comment telling the user how to run a GAMESS calculation (or generate a PUNCH file) with no symmetry.

tovrstra: @PaulWAyers Thanks for this. I was not aware of the symmetry-unique atoms. That would indeed make it hard at this stage to deal with anything else than C1.

FarnazH: I jsut added a comment to the code, and updated the raised message.

iodata/formats/gamess.py
FarnazH: Good point. Fixed it!

iodata/formats/gamess.py
```diff
@@ -0,0 +1,145 @@
...
+    counter = 0
+    while counter < natom:
+        words = next(lit).split()
+        for word in words:
+            masses[counter] = float(word)
+            counter += 1
+    return masses
+
+
+@document_load_one("PUNCH", ['title', 'energy', 'grot', 'atgradient', 'athessian', 'atmasses',
```
FarnazH, line 120: Just to double-check @tovrstra, should we keep this as `"PUNCH"` or change it to sth line `"DAT"` or something else?

iodata/formats/gamess.py
```diff
@@ -0,0 +1,145 @@
...
+        words = next(lit).split()
+        gradient[i] = words[2:5]
+    return energy, gradient
+
+
+def _read_hessian(lit: LineIterator, result: dict) -> np.ndarray:
+    """Extract ``hessian`` from the punch file."""
+    # check that $HESS is not already parsed
+    if "athessian" in result:
+        lit.error("Cannot parse $HESS twice! Make sure approximate hessian is not being parsed!")
```
FarnazH, line 90: I am not sure how to test this without adding a new test file (which is not desirable). Any suggestions @tovrstra?

iodata/formats/gamess.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _read_data(lit: LineIterator) -> tuple:
+    """Extract ``title``, ``symmetry`` and ``symbols`` from the punch file."""
+    title = next(lit).strip()
+    symmetry = next(lit).split()[0]
+    # The dat file only contains symmetry-unique atoms, so we would be incapable of
+    # supporting non-C1 symmetry without significant additional coding.
+    if symmetry != "C1":
+        raise NotImplementedError(f"Only C1 symmetry is supported. Got {symmetry}")
```
FarnazH, line 41: I am not sure how to test this without adding a new test file (which is not desirable). Any suggestions @tovrstra?

iodata/formats/gamess.py
tovrstra: PUNCH is fine.

iodata/formats/gamess.py
tovrstra: It is also impractical to add files at this stage. We would need to get in touch with someone who is at least a casual GAMESS user to generate good files. That would allow us to further elaborate on this format.

iodata/formats/gamess.py
tovrstra: See other comment. This is as good as we can get it at the moment.

PR #198, opened 2020-08-25
mwfn load_one

BradenDKelly:
Completed  load_one function support for *.mwfn file types

Separate future PR's will contain load_many, dump_one, dump_many

Files where added  in three locations:
    1) tests/data/ had the following files added for testing
        ch3_hf_sto3g_fchk_multiwfn3.7.mwfn
        ch3_rohf_sto3g_g03_fchk_multiwfn3.7.mwfn
        he_spdfgh_virtual_fchk_multiwfn3.7.,wfn
   2) tests/ had the following file added for testing:
        test_mwfn.py
    3) formats/ had the following file added for testing:
        mwfn.py


BradenDKelly:
@tovrstra @FarnazH Feel free to add yourselves as reviewers since I cannot.

tovrstra:
I'll fix some minor issues to make the tests pass before review.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+# https://chemrxiv.org/articles/Mwfn_A_Strict_Concise_and_Extensible_Format
+# _for_Electronic_Wavefunction_Storage_and_Exchange/11872524
+# For cartesian shells
+# S shell: S
+# P shell: X, Y, Z
+# D shell: XX, YY, ZZ, XY, XZ, YZ
+# F shell: XXX, YYY, ZZZ, XYY, XXY, XXZ, XZZ, YZZ, YYZ, XYZ
+# G shell: ZZZZ, YZZZ, YYZZ, YYYZ, YYYY, XZZZ, XYZZ, XYYZ, XYYY, XXZZ, XXYZ, XXYY, XXXZ, XXXY, XXXX
+# H shell: ZZZZZ, YZZZZ, YYZZZ, YYYZZ, YYYYZ, YYYYY, XZZZZ, XYZZZ, XYYZZ, XYYYZ, XYYYY, XXZZZ,
+#           XXYZZ, XXYYZ, XXYYY, XXXZZ, XXXYZ, XXXYY, XXXXZ, XXXXY, XXXXX
```
tovrstra, line 45: These differ from the HORTON2 conventions. Can you add them explicitly to the `CONVENTIONS` list below?

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+# D shell: D 0, D+1, D-1, D+2, D-2
+# F shell: F 0, F+1, F-1, F+2, F-2, F+3, F-3
+# G shell: G 0, G+1, G-1, G+2, G-2, G+3, G-3, G+4, G-4
+
+CONVENTIONS = {
+    (9, 'p'): HORTON2_CONVENTIONS[(9, 'p')],
+    (8, 'p'): HORTON2_CONVENTIONS[(8, 'p')],
+    (7, 'p'): HORTON2_CONVENTIONS[(7, 'p')],
+    (6, 'p'): HORTON2_CONVENTIONS[(6, 'p')],
+    (5, 'p'): HORTON2_CONVENTIONS[(5, 'p')],
```
tovrstra, line 56: I'd remove these if they are not defined for this format.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+    (0, 'c'): ['1'],
+    (1, 'c'): ['x', 'y', 'z'],
+    (2, 'c'): ['xx', 'yy', 'zz', 'xy', 'xz', 'yz'],
+    (3, 'c'): ['xxx', 'yyy', 'zzz', 'xyy', 'xxy', 'xxz', 'xzz', 'yzz', 'yyz', 'xyz'],
+    (4, 'c'): HORTON2_CONVENTIONS[(4, 'c')][::-1],
+    (5, 'c'): HORTON2_CONVENTIONS[(5, 'c')][::-1],
+    (6, 'c'): HORTON2_CONVENTIONS[(6, 'c')][::-1],
+    (7, 'c'): HORTON2_CONVENTIONS[(7, 'c')][::-1],
+    (8, 'c'): HORTON2_CONVENTIONS[(8, 'c')][::-1],
+    (9, 'c'): HORTON2_CONVENTIONS[(9, 'c')][::-1],
```
tovrstra, line 69: Idem, if not defined, better to leave them out. It will cause an error message when a file still contains them, which is a good reminder that we have to lookup and verify the conventions for higher angular momenta.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+                d[name] = int(line.split('=')[1].strip())
+                count += 1
+                break
+    return d['Nbasis'], d['Nindbasis'], d['Nprims'], d['Nshell'], d['Nprimshell']
+
+
+def _load_helper_atoms(lit: LineIterator, num_atoms: int) -> Tuple[np.ndarray, np.ndarray,
+                                                                   np.ndarray]:
+    """Read the coordinates of the atoms."""
+    atnums = np.empty(num_atoms, int)
```
tovrstra, line 110: We typically use `natom` for this in IOData. (It is a naming convention from the fortran world, which may seem odd, but it works well for keeping names short.) `s` is not included to reserve plural for arrays, and for brevity of course.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+    while '$Centers' not in line and line is not None:
+        line = next(lit)
+
+    for atom in range(num_atoms):
+        line = next(lit)
+        atnums[atom] = int(line.split()[2].strip())
+        atcorenums[atom] = float(line.split()[3].strip())
+        # extract atomic coordinates
+        coords = line.split()
+        atcoords[atom, :] = [coords[4], coords[5], coords[6]]
```
tovrstra, line 123: minor:
```suggestion
        atcoords[atom, :] = coords[4:7]
```

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+    atcorenums = np.empty(num_atoms, float)
+    atcoords = np.empty((num_atoms, 3), float)
+    line = next(lit)
+    while '$Centers' not in line and line is not None:
+        line = next(lit)
+
+    for atom in range(num_atoms):
+        line = next(lit)
+        atnums[atom] = int(line.split()[2].strip())
+        atcorenums[atom] = float(line.split()[3].strip())
```
tovrstra, line 120: Could you use `words = line.spit` upfront and use in all lines below? This is a bit nicer than splitting the same line several times.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+    -------------
+    shell_map:  np.ndarray (integer)
+        Index of what atom the shell is centered on. The mwfn file refers to this section
+        as `Shell centers`. Mwfn indices start at 1, this has been modified and starts
+        at 0 here. For water (O, H, H) with 6-31G, this would be an array like
+        [0, 0, 0, 0, 0, 1, 1, 2, 2]. , `O` in 6-31G has 5 shells and`H` has two shells.
+    shell_types: np.ndarray (integer)
+        Angular momentum of the shell. Indices start at 0 for 's' orbital, 1 for 'p' etc.
+        For 6-31G for a heavy atom this would be [0, 0, 1, 0, 1] corresponding
+         to [1s, 2s, 2p, 2s, 2p]
```
tovrstra, line 271: This may break sphinx formatting:
```suggestion
        to [1s, 2s, 2p, 2s, 2p]
```

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+
+def build_obasis(shell_map: np.ndarray, shell_types: np.ndarray,
+                 exponents: np.ndarray, prim_per_shell: np.ndarray,
+                 coeffs: np.ndarray,
+                 ) -> Tuple[MolecularBasis]:
+    """Based on the fchk modules basis building.
+
+    Parameters
+    -------------
+    shell_map:  np.ndarray (integer)
```
tovrstra, line 263: In principle, no need to document types here when type hints are used, but I see your point that type hints do not contain the array dtype. So, this is just fine. You may want to vote here: https://github.com/numpy/numpy/issues/7370

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+            'prim_per_shell': prim_per_shell, 'exponents': exponent, 'coeffs': coeffs,
+            'mo_numbers': mo_numbers, 'mo_occs': mo_occs, 'mo_energies': mo_energies,
+            'mo_coeffs': mo_coeffs, 'mo_type': mo_type, 'mo_sym': mo_sym}
+
+
+def build_obasis(shell_map: np.ndarray, shell_types: np.ndarray,
+                 exponents: np.ndarray, prim_per_shell: np.ndarray,
+                 coeffs: np.ndarray,
+                 ) -> Tuple[MolecularBasis]:
+    """Based on the fchk modules basis building.
```
tovrstra, line 259: Could you make the first line describe the function?

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+
+@document_load_one("MWFN", ['atcoords', 'atnums', 'atcorenums', 'energy',
+                            'mo', 'obasis', 'extra', 'title'])
+def load_one(lit: LineIterator) -> dict:
+    """Do not edit this docstring. It will be overwritten."""
+    inp = load_mwfn_low(lit)
+
+    # MWFN contains more information than most formats, so the following dict
+    # stores some "extra" stuff.
+    mwfn_dict = {
```
tovrstra, line 311: Could you use `extra` instead of `mwfn_dict` for clarity. Some things in this dictionary are also present in the other IOData attributes, either directly or indirectly. I'd suggest to take these out. See line 13 of `import this`. :)

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+        Array of same length as `exponents` containing orbital expansion coefficients.
+    """
+    shells = []
+    counter = 0
+    # First loop over all shells
+    for i, n in enumerate(prim_per_shell):
+        shells.append(Shell(
+            shell_map[i],
+            [abs(shell_types[i])],
+            ['p' if shell_types[i] < 0 else 'c'],
```
tovrstra, line 288: This seems to contradict the conventions for pure functions mentioned above?

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+            'atnums': atnums, 'atcoords': atcoords, 'atcorenums': atcorenums,
+            'nbasis': nbasis, 'nindbasis': nindbasis, 'nprims': nprim,
+            'nshells': nshell, 'nprimshells': nprimshell, 'full_virial_ratio': vt_ratio,
+            'shell_centers': shell_centers, 'shell_types': shell_types,
+            'prim_per_shell': prim_per_shell, 'exponents': exponent, 'coeffs': coeffs,
+            'mo_numbers': mo_numbers, 'mo_occs': mo_occs, 'mo_energies': mo_energies,
+            'mo_coeffs': mo_coeffs, 'mo_type': mo_type, 'mo_sym': mo_sym}
+
+
+def build_obasis(shell_map: np.ndarray, shell_types: np.ndarray,
```
tovrstra, line 255: unless you think this might be useful for other formats:
```suggestion
def _build_obasis(shell_map: np.ndarray, shell_types: np.ndarray,
```

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+                           inp['mo_energies'],
+                           None,
+                           )
+
+    return {
+        'title': inp['title'],
+        'atcoords': inp['atcoords'],
+        'atnums': inp['atnums'],
+        'atcorenums': inp['atcorenums'],
+        'charge': inp['charge'],
```
tovrstra, line 361: Charge can be left out. I will be derived from `atcorenums` and `mo`.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+        'obasis': obasis,
+        'mo': mo,
+        'nelec': inp['nelec_a'] + inp['nelec_b'],
```
tovrstra, line 364: `nelec` can also be left out for the same reason.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,210 @@
...
+def load_fchk_helper(fn_fchk):
+    """Load a testing fchk file with iodata.iodata.load_one."""
+    with path('iodata.test.data', fn_fchk) as fn:
+        return load_one(fn)
+
+
+def test_load_mwfn_nonexistent():
+    with pytest.raises(IOError):
+        with path('iodata.test.data', 'fubar_crap.fchk') as fn:
+            load_one(str(fn))
```
tovrstra, line 47: Can probably be omitted.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,210 @@
...
+    assert_equal(mol.extra['mo_type'][0], 0)
+    assert_equal(mol.extra['mo_sym'][0], '?')
+    # test that for the same molecule fchk and mwfn generate the same objects.
+    olp = compute_overlap(mol.obasis, mol.atcoords)
+    mol2 = load_fchk_helper('ch3_rohf_sto3g_g03.fchk')
+    olp_fchk = compute_overlap(mol2.obasis, mol2.atcoords)
+    # assert_allclose fails due to abs/rel tolerance settings
+    assert np.allclose(mol.atcoords, mol2.atcoords)
+    assert_allclose(mol2.obasis.shells[0].coeffs, coeffs1)
+    # fails due to special SP basis storage (leaving commented lines in for future reference)
```
tovrstra, line 115: This can still be tested if you like, not 100% sure though. Can you try `mol2.obasis.get_segmented`?

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,210 @@
...
+                            3.28562907E-01, 7.04456296E-01, 7.88139770E-01, 7.89228899E-01])
+    assert_allclose(mol.mo.energies, mo_energies)
+    assert_equal(mol.mo.occs[0], 1.000000)
+    assert_equal(mol.extra['mo_type'][0], 1)
+    assert_equal(mol.extra['mo_sym'][0], '?')
+    # test that for the same molecule fchk and mwfn generate the same objects.
+    olp = compute_overlap(mol.obasis, mol.atcoords)
+    mol2 = load_fchk_helper('ch3_hf_sto3g.fchk')
+    olp_fchk = compute_overlap(mol2.obasis, mol2.atcoords)
+    # assert_allclose fails due to abs/rel tolerance settings
```
tovrstra, line 144: `assert_allclose` has optional arguments to change atol and rtol. I'm not sure if it matters a lot. I believe assert_allclose gives nices error messages upon failures.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,210 @@
...
+    mol2 = load_fchk_helper('he_spdfgh_virtual_fchk_multiwfn3.7.mwfn')
+    assert mol2.nelec == 2
+    assert mol2.charge == 0
+    assert mol2.extra['nelec_a'] == 1
+    assert mol2.extra['nelec_b'] == 1
+    mol3 = load_fchk_helper('ch3_hf_sto3g_fchk_multiwfn3.7.mwfn')
+    assert mol3.nelec == 9
+    assert mol3.charge == 0
+    assert mol3.extra['nelec_a'] == 5
+    assert mol3.extra['nelec_b'] == 4
```
tovrstra, line 163: These may be affected by #196

tovrstra (commented):
I have mostly minor comments. I'm only a bit puzzled that the code seems to assume Cartesian functions in `build_obasis`.

iodata/test/test_mwfn.py
BradenDKelly: I just realized that mol2 and mol3 use the load_fchk_helper rather than the load_mwfn_helper. Obviously, one of the helpers is superfluous since they use load_one from the api, which bases its decisions on the format. I will remove one of the two and give the helper function a generic name.

iodata/test/test_mwfn.py
BradenDKelly: I can make this change. It was one based on momentary convenience, which is not at the top of the pyramid for design choices.

iodata/test/test_mwfn.py
BradenDKelly: removed.

iodata/formats/mwfn.py
BradenDKelly: The use of the build_obasis was an effort to follow the wfn.py format as a template. The actual code inside it is based on fchk.py, found on line 131 of fchk.py, modified for the differences found in mwfn. I doubt the build_obasis will be generally useful, and will rename to _build_obasis.

iodata/formats/mwfn.py
BradenDKelly: I will review. As noted in next comment, I was building this based on wfn.py and fchk.py and this may not be as robust as hoped.

iodata/formats/mwfn.py
BradenDKelly: I also use natom in my fortran. this change makes sense to me.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+
+    for atom in range(num_atoms):
+        line = next(lit)
+        atnums[atom] = int(line.split()[2].strip())
+        atcorenums[atom] = float(line.split()[3].strip())
+        # extract atomic coordinates
+        coords = line.split()
+        atcoords[atom, :] = [coords[4], coords[5], coords[6]]
+        # return but convert angstroms to amu
+    return atnums, atcorenums, atcoords / spc.value(u'atomic unit of length') / 1E10
```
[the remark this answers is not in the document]
BradenDKelly: yes, replacing with atcoords * angstroms. I would like to note that I find this confusing however. It would be nicer if iodata.utils defined this as ang_to_amu or something that is explicitly a conversion from one unit to another. While one knows angstroms are involved in atcoords * angstoms, it is not clear what the conversion unit to/from is.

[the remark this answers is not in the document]
BradenDKelly: Given that iodata explicitly states it uses atomic units, perhaps my above comment if a bit narrow viewed, and upon remembering that iodata specifically has default units, all conversions must be to/from them.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+
+import numpy as np
+import scipy.constants as spc
+
+from ..basis import HORTON2_CONVENTIONS, MolecularBasis, Shell
+from ..docstrings import document_load_one
+from ..orbitals import MolecularOrbitals
+from ..utils import LineIterator
+
+__all__ = ['load_mwfn_low']
```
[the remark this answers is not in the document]
BradenDKelly: I forsaw this one. Pycharm was not happy with [], but, Travis is more important.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,367 @@
...
+    mo_type = int(next(lit).split()[1])
+    energy = float(next(lit).split()[1])
+    occ = float(next(lit).split()[1])
+    sym = str(next(lit).split()[1])
+    next(lit)  # skip line
+    coeffs = _load_helper_section(lit, nbasis, '', 0, float)
+    return number, occ, energy, coeffs, mo_type, sym
+
+
+def load_mwfn_low(lit: LineIterator) -> dict:
```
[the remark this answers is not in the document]
BradenDKelly: will do.

iodata/formats/mwfn.py
BradenDKelly: will do.

iodata/formats/mwfn.py
BradenDKelly: Old habits from fortran. I like knowing all types beforehand.

iodata/formats/mwfn.py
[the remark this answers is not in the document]
tovrstra: Indeed, the `ang_to_bohr` style would be ok if the "to" part were not be fixed by convention.

The following makes it easy to remember how it works:

- When you say "this bond length is 1.5 Å", the python equivalent is `bond_length = 1.5 * angstrom`.

- The other way is often found in graphs in old papers, where the axis labels to indicate a bond length in angstrom is printed as "Bond length / Å". Expressing this in python would be `print("Bond length in Angstrom:", bond_length / angstrom)`

Should we add this explanation to the docs somewhere? It might be helpful.

iodata/formats/mwfn.py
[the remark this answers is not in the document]
tovrstra: That's why I'm using a bare bones text editor. PyCharm is the Python-equivalent of MS Word trying to tell me how to write "IOData" as "Iodata".

iodata/formats/mwfn.py
BradenDKelly: Actually, this is for atomic numbers, i.e., oxygen is 8, carbon is 6. I have changed the variable to `atomic_numbers` to provide a better description. Is this alright?

iodata/formats/mwfn.py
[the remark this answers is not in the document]
BradenDKelly: This one seems to be tied in with the outstanding travis error

iodata/formats/mwfn.py
tovrstra: I'm sorry for the confusion. What I meant is `num_atoms` -> `natom`.

It would be good to use the names from the ``IOData`` class, see https://github.com/theochem/iodata/blob/master/iodata/iodata.py#L64 (We've discussed these upfront some time ago and tried to find compromises between brevity and clarity.) In general, reusing names from the iodata modules like `iodata.iodata`, `iodata.api`, etc. results in consistent names, which makes the code easier to understand.

iodata/formats/mwfn.py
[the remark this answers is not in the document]
BradenDKelly: It can't hurt although I was more after the fact that only one unit is known. But, I had to correct myself because IOData always uses bohr for its length units, so if you see a conversion of lengths, you know it is going from Bohr to whatever units is given, e.g., angstrom.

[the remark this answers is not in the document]
tovrstra: ok, thanks.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,315 @@
...
+
+    # skip lines until "$Centers" section is reached
+    line = next(lit)
+    while '$Centers' not in line and line is not None:
+        line = next(lit)
+
+    for atom in range(natom):
+        words = next(lit).split()
+        data["atnums"][atom] = int(words[2].strip())
+        data["atcorenums"][atom] = float(words[3].strip())
```
tovrstra, line 136: very minor:
```suggestion
        data["atnums"][atom] = words[2]
        data["atcorenums"][atom] = words[3]
```
Split returns stripped results and numpy does conversion from string upon assignment. E.g.

```python
>>> import numpy as np
>>> a = np.zeros(3)
>>> a[1] = "5.0"
>>> a
array([0., 5., 0.])
>>> b = np.zeros(3, int)
>>> b[0] = "134 "
>>> b
array([134,   0,   0])
```

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,315 @@
...
+def _load_helper_section(lit: LineIterator, nprim: int, start: str, skip: int,
+                         dtype: np.dtype) -> np.ndarray:
+    """Read single or multiple line(s) section."""
+    section = []
+    while len(section) < nprim:
+        line = next(lit)
+        if not line.startswith(start):
+            lit.error(f"Expected line to start with {start}! Got line={line}")
+        section.extend(line.split()[skip:])
+    return np.array(section).astype(dtype)
```
tovrstra, line 177:
```suggestion
    return np.array(section, dtype)
```

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,315 @@
...
+    for index in range(n_mo):
+        line = next(lit)
+        while 'Index' not in line:
+            line = next(lit)
+        assert line.startswith('Index')
+        data["mo_numbers"][index] = int(line.split()[1])
+        data["mo_type"][index] = int(next(lit).split()[1])
+        data["mo_energies"][index] = float(next(lit).split()[1])
+        data["mo_occs"][index] = float(next(lit).split()[1])
+        data["mo_sym"][index] = str(next(lit).split()[1])
```
tovrstra, line 200: Type conversions can be omitted.

iodata/formats/mwfn.py
```diff
@@ -0,0 +1,315 @@
...
+    # get number of basis & molecular orbitals (MO)
+    # Note: MWFN includes virtual orbitals, so num_mo equals number independent basis functions
+    num_basis = data["Nindbasis"]
+    num_mo = data["Nindbasis"]
+    if data["mo_kind"] == "unrestricted":
+        num_mo *= 2
+    # load MO information
+    data.update(_load_helper_mo(lit, num_basis, num_mo))
+
+    # TODO: load data in '# Various matrices' section
```
tovrstra, line 258: Should we open an issue for this? I'm worried that leaving it here as a TODO item will not be noticed.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,196 @@
...
+from ..utils import LineIterator
+
+try:
+    from importlib_resources import path
+except ImportError:
+    from importlib.resources import path
+
+
+def helper_load_mwfn_low(fn_mwfn):
+    """Load a testing Gaussian log file with iodata.formats.mwfn.load_mwfn_low."""
```
tovrstra, line 36: The docstring may need to be fixed. The "Gaussian" part is probably inspired by another format.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,196 @@
...
+    with path('iodata.test.data', fn_mwfn) as fn:
+        lit = LineIterator(str(fn))
+        return _load_mwfn_low(lit)
+
+
+def load_format_helper(fn_format):
+    """Load a testing formatted file with iodata.iodata.load_one."""
```
tovrstra, line 43: idem, with the word checkpoint removed.

iodata/test/test_mwfn.py
```diff
@@ -0,0 +1,196 @@
...
+    assert_equal(mol.extra['mo_sym'][0], '?')
+    # test that for the same molecule fchk and mwfn generate the same objects.
+    olp = compute_overlap(mol.obasis, mol.atcoords)
+    mol2 = load_format_helper('ch3_rohf_sto3g_g03.fchk')
+    olp_fchk = compute_overlap(mol2.obasis, mol2.atcoords)
+    assert_allclose(mol.atcoords, mol2.atcoords, atol=1E-7, rtol=1E-7)
+    assert_allclose(mol2.obasis.shells[0].coeffs, coeffs1)
+    # fails due to special SP basis storage (leaving commented lines in for future reference)
+    # assert_allclose(mol2.obasis.shells[1].coeffs, coeffs2)
+    # assert_allclose(mol2.obasis.shells[2].coeffs, coeffs3)
```
tovrstra, line 108: This should be fixable. Something like:
```suggestion
    # Mind the SP contraction...
    assert_allclose(mol2.obasis.shells[1].coeffs[:, 0], coeffs2)
    assert_allclose(mol2.obasis.shells[1].coeffs[:, 1], coeffs3)
    assert_allclose(mol2.obasis.shells[2].coeffs, coeffs4)
```
(may need to be further corrected though.)

tovrstra (commented):
I have a few very minor comments, which should be easy to fix. Thanks for all the nice work!

iodata/test/test_mwfn.py
BradenDKelly: assert_allclose(mol2.obasis.shells[1].coeffs[:, 0], np.squeeze(coeffs2.T)) works well enough

iodata/test/test_mwfn.py
BradenDKelly: This function appeared but is not used in the testing anywhere. Possible in preparation for a future use? @FarnazH

iodata/test/test_mwfn.py
BradenDKelly: I don't quite understand what needs to be done here :)

iodata/formats/mwfn.py
BradenDKelly: I had the same TODO, but worded differently. The issue I had was that the file is read line by line, and so when scanning for the optional keywords, I would hit end of file and get an error, if they weren't there. I think the fix is probably simple. A brute force option would be to open the file, scan or optional keywords, then proceed with normal scan. An another would be to have the optional keywords passed by caller. I feel however, that there is a simple fix to the end-of-line error.

iodata/test/test_mwfn.py
FarnazH: If I am not wrong, this function was added by you @BradenDKelly, so probably it's leftover from earlier code? I agree with you, it is not used, so it's better to be removed. Thanks.

iodata/formats/mwfn.py
FarnazH: Alternatively, to avoid searching for optional keywords (and going through the file twice), we can have a list of all possible keywords that appear in `mwfn` files, and use a `while` loop to properly parse what is in a file (like 'fchk')... This would be more robust and easier to extend. 

Currently we are relying a specific order of data in `mwfn` file, which would not be problematic as only `Multiwfn` writes `mwfn` files, and as @BradenDKelly commented the order of entries must be fixed. The `'# Various matrices'` section includes alpha/beta/total density matrix, overlap matrix, etc. (we can get a list of these from the paper but it is meant to be a place for dumping matrices used for post processing). As this PR is getting long, I think it's best to open an issue for paring `'# Various matrices'` section of mwfn @tovrstra.

iodata/test/test_mwfn.py
tovrstra: The docstring seems to inconsistent. What is the word "formatted" referring to. I guess it is a left-over from "formatted checkpoint file"

tovrstra:
@BradenDKelly I've merged master back into this branch to run it through the rebooted CI. I'll go through it later. This has been open for too long.

iodata/formats/mwfn.py
tovrstra: I'll make the issue and remove the TODO line, because it may trip pylint (depends on local config).

iodata/formats/mwfn.py
tovrstra: It was actually OK. I just misunderstood the code.

iodata/test/test_mwfn.py
tovrstra: I'll take care of this.

tovrstra:
@BradenDKelly @FarnazH  I fixed a few minor issues. As far as I'm concerned, this is ready to be merged.

@BradenDKelly You may want to double check my changes in the initialization of the Orbitals object. I believe there was a mistake with the `norba` and `norbb` arguments. I've updated their docstrings to clarify their meaning.

tovrstra (dismissed):

FarnazH:
Thanks for fixing the mistake in initializing `MolecularOrbitals` @tovrstra. I think the new validators added in issue #245 would have caught it, yes?
 
I took a last look and removed redundant properties stored in the `extra` dictionary. I am good with merging this PR.

FarnazH (approved):

FarnazH (approved):
Thanks for fixing the mistake in initializing `MolecularOrbitals` @tovrstra. I think the new validators added in issue #245 would have caught it, yes?
 
I took a last look and removed redundant properties stored in the `extra` dictionary. I am good with merging this PR.

tovrstra:
Thanks @FarnazH. I'm good with merging as it is now. @BradenDKelly If you notice something after merging, feel free to open another PR. In any case, thanks everyone for the nice work.

BradenDKelly:
Thanks @tovrstra and @FarnazH. I have a prototype of a dump_one I will whittle away at. Also, somewhat of a tangent, but it does not appear that we have a *.cif reader yet do we? Thinking selfishly of future Braden, I think if we do not, I will open a PR soon for that and get a load_one, dump_one for *.cif, or at least, a load_one.

tovrstra:
Let's open an issue for CIF, where we can collect useful info regarding this format. It is not entirely trivial but certainly useful. The format is fairly complicated, with it's own container format and a specification within that container.

PaulWAyers:
@BradenDKelly if you polish your prototype of dump_one for *.mwfn please make sure you go through the load_one to make sure whatever changes Toon and Farnaz made there filter down....less work for everyone that way.

PR #199, opened 2020-08-26
Rename pip package with qc prefix

### LINKED ISSUE #185: Document installation of pre-releases
The current install documentation only explains stable releases and we don't have any serious stable releases yet.

tovrstra:
Fixes #185. (This is also to avoid name collisions on pypi.)

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -33,7 +33,9 @@ To install IOData using conda package management system, install
 
 .. code-block:: bash
 
-    # Activate your main conda environment if needed.
+    # Activate your main conda environment if it is not loaded in your .bashrc.
+    # E.g. run the following if you have miniconda installed in e.g. ~/miniconda3
+    source ~/miniconda3/bin/activate
 
     # Create a horton3 conda environment. (optional, recommended)
     conda create -n horton3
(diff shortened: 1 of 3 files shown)


tovrstra:
This works as expected. I'm going to leave the PR open and also update the install documentation accordingly.

tovrstra:
You can see the yanked release here, which I used for testing: https://pypi.org/project/qc-iodata/1.0.0a1.post80/

tovrstra:
Few comments on the changes:
- I've stripped the `$` signs from tha bash commands to facilitate copy-pasting.
- Instructions for development version are only put in the docs to avoid confusion.

setup.py
```diff
@@ -50,7 +50,7 @@ def get_readme():
 VERSION, DEV_CLASSIFIER = get_version_info()
 
 setup(
-    name='iodata',
+    name='qc-iodata',
```
evohringer, line 53: Do we need to change the name here? I´m asking because the package name would be maintained as iodata, right?

doc/install.rst
```diff
@@ -27,21 +27,37 @@ Installation
...
+To install IOData with pip, you may want to create a `virtual environment`_,
+and then:
 
 .. code-block:: bash
 
-    pip install iodata
-    pip install iodata --user
-    python3 -m pip install iodata
-    python3 -m pip install iodata --user
+    # Install the stable release.
```
evohringer, line 56: I would suggest leaving the option with --user as an extra option not depending explicitly on having a conda environment and installing as a user in the home directory. Not all users might be aware of this option.

setup.py
tovrstra: Yes, the only serious role of this string is to determine the name of the package on https://pypi.org/. It also affects the filenames of distribution files generated by `setup.py`. The package name for importing remains the same. `import iodata` still works as usual.

doc/install.rst
tovrstra: Agreed. Also the `python3 -m pip install` should be mentioned somehow. That one can be a workaround when the PATH variable is incomplete, which would be nice to explain.

tovrstra:
I've cleaned up a few more things. The list of dependencies was removed. It is not so essential because package managers take care of this. (It was also outdated.)

doc/install.rst
tovrstra: This is fixed.

tovrstra:
@FarnazH Thanks for the comments.

Regarding the dependencies, my reasoning was that normal users do not need to know this information explicitly. Pip and conda take care of installing these when the user installs IOData. If someone deliberately does not want to use pip or conda for some technical reason, I assume such a person is sufficiently skilled to check `setup.py`. That said, I think we can still mention the dependencies in the install docs and remark that these do not need to be installed manually.

I agree with the links. These would be useful to have. (One can click on the badges, but this is not so well known.)

doc/install.rst
```diff
@@ -24,32 +24,129 @@
...
+       # (Only do this if you understand the implications.)
+       pip3 install --pre qc-iodata
+       # alternative: python3 -m pip install --pre qc-iodata
+
+2. You can install into your ``${HOME}`` directory, without creating a virtual
+   environment.
+
+   .. code-block:: bash
+
+       # Install the sable release in your home directory.
```
lmacaya, line 95:
```suggestion
       # Install the stable release in your home directory.
```

tovrstra: thanks, fixed it.

tovrstra:
Thanks for reviewing everyone. I'm going to merge so we can make a proper alpha release soon. (If needed, further fixes to the install docs are always welcome.)

PR #202, opened 2020-08-26
Update run_type definition in IOData docstring

### LINKED ISSUE #162: Specify `runtype` vs. `run_type` and determine valid values
In #41 the attribute name storing the calculation type is given as `runtype`. Currently it is implemented as `run_type` in the fchk format parser and the `IOData` class. It would likely be easier for @tovrstra to edit this in the master list on #41 than to change the API now (I don't prefer one name over the other too strongly). 

In the fchk parser, the [current possible values](https://github.com/theochem/iodata/blob/master/iodata/formats/fchk.py#L115) for `run_type` are `'energy'`, `'opt'`, `'scan'`, and `'freq'`; in the `IOData` class docs, the [values given as examples](https://github.com/theochem/iodata/blob/master/iodata/iodata.py#L171) are `'energy'`, `'energy_force'`, `'opt'`, and `'freq'`. Should we be trying to make an exhaustive list of the possible `run_type` values now, or add them as they come? Regardless, the possible values should be documented (likely in the class docs), so that we're not ending up with redundant options.

This issue is important for generating input files (#43) and dumping input files to QCSchema or any other format (#39).

tovrstra:
Fixes #162.

### DIFF
--- a/iodata/iodata.py
+++ b/iodata/iodata.py
@@ -169,8 +169,9 @@ class IOData:
         matrices. Names can be ``scf``, ``post_scf``, ``scf_spin``,
         ``post_scf_spin``. These matrices are always expressed in the AO basis.
     run_type
-        The type of calculation that lead to the results stored in IOData, e.g.
-        'energy', 'energy_force', 'opt', 'freq', ...
+        The type of calculation that lead to the results stored in IOData, which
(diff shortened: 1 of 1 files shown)


wilhadams (approved):
Looks good to me!

wilhadams:
Closes #162

PR #205, opened 2020-08-27
Minor fixes in GROMACS and CHARMM documentation.

lmacaya:
This PR is just for minor changes. I modified the format of the documentation, in a more consistent way with respect to the other formats.

### DIFF
--- a/iodata/formats/charmm.py
+++ b/iodata/formats/charmm.py
@@ -16,25 +16,16 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, see <http://www.gnu.org/licenses/>
 # --
-"""CHARMM coordinate (crd) file format.
+"""CHARMM crd file format.
 
-CHARMM coordinate files contain information about the location of each atom in Cartesian (3D) space.
-The format of the ASCII (CARD) CHARMM coordinate files is:
-    * Title line(s)
-    * Number of atoms in file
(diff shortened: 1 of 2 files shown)


iodata/formats/charmm.py
```diff
@@ -16,25 +16,16 @@
...
-    * X-coordinate
-    * Y-coordinate
-    * Z-coordinate
-    * Segment identifier
-    * Residue identifier
-    * Weighting array value
+The coordinate lines contain specific information about each atom.
+These have the following structure: Atom number (sequential), residue number
+(specified relative to first residue in the PSF), residue name, atom type, x-coordinate,
+y-coordinate, z-coordinate, segment identifier, residue identifier and a weighting array value.
```
FarnazH, line 28: The current documentation does not look good, so hopefully this fixes it. See:
https://iodata.readthedocs.io/en/latest/pyapi/iodata.formats.charmm.html#module-iodata.formats.charmm

tovrstra: For quick RST tests, you can use the online editor: http://rst.ninjs.org/ This is much faster than locally building docs.

Normally in RST, a bullet list should be separated from the preceding paragraph by an empty line. Indentation is not needed, e.g.

```
Test sentence:

* item 1
* item 2
```

Anyway, this is good too.

PR #206, opened 2020-08-29
Fix Failure of Parsing PDB wihout an END

FarnazH:
I am not sure how common (or even legitimate) this issue is, but I have been encountering PDB files that don't have an `END`, so trying to load them fails... I can investigate more to see exactly when such PDB files are found, but I think it was when using Biopython to manipulate a PDB file...

I changed the `load_one` to print a warning message if `END` is not found, and still return the data. I have attached an example, but didn't add it to the test files, because I wasn't sure it is worth it. We can remove the `END` of one of the existing pdb files to test this, probably.
[1JBEA.pdb.tar.gz](https://github.com/theochem/iodata/files/5144873/1JBEA.pdb.tar.gz)

Please feel free to let me know what you think.


tovrstra:
@FarnazH I agree. To my knowledge, such a missing END in a PDB file is so common that not everyone thinks of them as errors. (It should be there, see http://www.wwpdb.org/documentation/file-format-content/format33/sect11.html#END)

tovrstra (approved):
No comments. A warning is indeed in order because, with the missing END, one cannot make the distinction between a file that fails to follow the standard, or one that is just incomplete because of an interrupted download.

tovrstra:
I'm trying to get an overview of the possible cases for the test files of MDTraj. It is a mess, but there seem to be three possible cases:

- single frame without `END`: https://github.com/mdtraj/mdtraj/blob/master/tests/data/1bpi.pdb and https://github.com/mdtraj/mdtraj/blob/master/tests/data/2waters_baker_hubbard.pdb
- trajectory with multiple `END` markers: https://github.com/mdtraj/mdtraj/blob/master/tests/data/2lines.pdb
- trajectory with `MODEL` and `ENDMDL` for frames and one final `END`: https://github.com/mdtraj/mdtraj/blob/master/tests/data/2EQQ.pdb

There are plenty more in that directory.

tovrstra (approved):
I've added more tests for all the variants and fixed minor snag in the unit tests. All is good.

tovrstra:
Thanks everyone! I'll merge.

PR #207, opened 2020-08-29
Fix Failure in Parsing Q-Chem Fchk

FarnazH:
The FCHK files generated by Q-Chem 5.2 & 5.3 does not contain "Total Energy", so parsing them with IOData fails. The code is changed to return `None` for `energy`, if "Total Energy" is not present in FCHK file; as a result, `energy` becomes an ifpresent value. 

Here are examples of FCHK generated by Q-Chem 5.2 & 5.3:
[Water_qchem5.2.fchk.tar.gz](https://github.com/theochem/iodata/files/5144896/Water_qchem5.2.fchk.tar.gz)
[Water_qchem5.3.fchk.tar.gz](https://github.com/theochem/iodata/files/5144897/Water_qchem5.3.fchk.tar.gz)

I can generate any test file required, so please feel free to let me know what you think.


tovrstra:
Looks good, I'd suggest to add a small (STO-3G water) fchk file to the tests to make sure it keeps working in future.

FarnazH:
Sure, I will do that @tovrstra. If we are adding a Q-Chem FCHK test, does it make sense to include higher order angular momentum too, or FCHK files are robust across various softwares? I am asking because I am generating some Q-Chem Molden files for testing like Zn with pure/Cartesian cc-pvqz basis (following the existing examples).

tovrstra:
@FarnazH That's a good point. You only know for sure if it works when it is tested.

evohringer (dismissed):
Looks fine!

evohringer:
One additional comment: The missing energy field also happens in molden files if I remember correctly. It might be an issue when using e.g. chemtools.

FarnazH:
I will add a test fchk file from Q-Chem soon.
I agree @evohringer, we need to make sure proper error is raised if a user tries to use such files for computing certain descriptors with chemtools.

tovrstra:
@FarnazH It would be nice to pick this one up again. Could you generate a smaller FCHK file with Q-Chem? The files you showed are relatively large. Thanks.

FarnazH:
@tovrstra I used a small calculation, but feel free to let me know if any other test files are needed. I can easily generate them (I will be making a PR with a Q-Chem MOLDEN file soon).

tovrstra (approved):
Looks very good! Thanks.

PR #208, opened 2020-08-30
Enable coverage upload to codecov in Roberto

tovrstra:
I'm not sure why this was not enabled yet. It is helpful for reviewing.


tovrstra:
@FarnazH I'm going to merge this (without formal review) to enable coverage analysis on our PRs. Coverage checks may not always be accurate, so I will not (yet) make them required. They are just helpful to detect issues or missing tests that are otherwise easily overlooked

PR #209, opened 2020-08-30
Update qchemlog.py to load EDA jobs

leila-pujal:
Add and modify code in qchemlog.py to be able to load information from Energy Decomposition Analysis qchemlog jobs, with a new _helper_eda() function.


iodata/formats/qchemlog.py
```diff
@@ -366,3 +397,71 @@ def _helper_thermo(lit: LineIterator) -> Tuple:
...
+                if line_2.startswith('     --------------------'):
+                    break
+                else:
+                    info = line_2.strip().split()
+                    if info[0] == 'E_elec':
+                        eda2_dic[info[0].lower()] = float(info[4])
+                    elif info[0] == 'E_pauli':
+                        eda2_dic[info[0].lower()] = float(info[4])
+                    elif info[0] == 'E_disp':
+                        eda2_dic[info[0].lower()] = float(info[4])
```
FarnazH, line 419:
```suggestion
                    if info[0] in ['E_elec', 'E_pauli', 'E_disp']:
                        eda2_dic[info[0].lower()] = float(info[4])
```

iodata/formats/qchemlog.py
```diff
@@ -366,3 +397,71 @@ def _helper_thermo(lit: LineIterator) -> Tuple:
...
+                    elif info[0].split("[")[-1] == 'PAULI':
+                        eda2_dic[info[0].split("[")[-1].lower()] = float(info[1].split("]")[0])
+                    if info[0] == 'DISPERSION':
+                        eda2_dic[info[0].lower()] = float(info[1])
+                    elif info[0] == 'POLARIZATION':
+                        eda2_dic[info[0].lower()] = float(info[1])
+                    elif info[0] == 'CHARGE':
+                        eda2_dic[info[0].lower() + ' ' + info[1].lower()] = float(info[2])
+                    elif info[0] == 'TOTAL':
+                        eda2_dic[info[0].lower()] = float(info[1])
```
FarnazH, line 463: I think the same type of change (mentioned above), can make this shorter...

iodata/formats/qchemlog.py
```diff
@@ -125,20 +131,39 @@ def load_qchemlog_low(lit: LineIterator) -> dict:
...
             data.update(_helper_job(lit))
         # standard nuclear orientation
-        elif line.startswith('Standard Nuclear Orientation (Angstroms)'):
+        elif line.startswith('Standard Nuclear Orientation (Angstroms)') and 'atcoords' not in data:
             # atnums, alpha_elec, beta_elec, nbasis, nuclear_replusion_energy, energy, atcoords
-            _, data['alpha_elec'], data['beta_elec'], data['nbasis'], \
-                data['nuclear_repulsion_energy'], data['energy'], _ = _helper_electron(lit)
+            data['atnums'], data['alpha_elec'], data['beta_elec'], data['nbasis'], \
+                data['nuclear_repulsion_energy'], data['atcoords'] = _helper_electron(lit)
+            data['natom'] = len(data['atnums'])
```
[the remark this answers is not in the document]
leila-pujal: I tried to remove it but it is used in  `_helper_hessian(lit, data['natom'])`

iodata/formats/qchemlog.py
```diff
@@ -125,20 +131,39 @@ def load_qchemlog_low(lit: LineIterator) -> dict:
 
         # get the atomic information
         if line.startswith('$molecule'):
-            result = _helper_atoms(lit)
-            data['charge'], data['spin_multi'], data['atnums'], data['atcoords'] = result
-            data['natom'] = len(data['atnums'])
+            # net charge and spin multiplicity
+            data['charge'], data['spin_multi'] = [int(i) for i in next(lit).strip().split()]
```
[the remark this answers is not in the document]
leila-pujal: You are right. I deleted the whole if section as `spin_mult` is saved in extras and the iodata attribute `spinpol` can't be set if `mo` is present. For charge if it is none in iodata object is derived from `self.atcorenums.sum() - self.nelec`

iodata/formats/qchemlog.py
leila-pujal: I couldn't do it for all `elif` sections because some are specific for that particular component but I think it reduced length quite a bit. Thanks for the suggestion!

iodata/formats/qchemlog.py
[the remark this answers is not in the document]
FarnazH: That's great. Thanks.

iodata/formats/qchemlog.py
```diff
@@ -56,7 +56,7 @@ def load_one(lit: LineIterator) -> dict:
     # build molecular orbitals
     # ------------------------
     # restricted case
-    if not data['unrestricted']:
+    if data['unrestricted'] is False:
```
tovrstra, line 59: I believe this change goes against PEP8. See https://www.python.org/dev/peps/pep-0008/ (Look for `Don't compare boolean values to True or False`) Is there is a special reason for this change?

I'll take a closer look at the whole PR soon.

leila-pujal: Not any reason that I can recall now or that I have in my notes....probably I change it because it felt more "correct"  in my naive python knowledge, although according to PEP8 is the worst of possible options... I am going to change it as well as two more I've seen and commit the changes. Thanks for noticing!

tovrstra:
I noticed a few very minor things, which I fixed right away. I'll wait for the tests to pass and then merge. Thanks everyone for all the nice work.

tovrstra (dismissed):

PR #211, opened 2020-09-02
Fix codecov configuration

tovrstra:
This should disable bogus commit failures due to false coverage issues.


tovrstra:
@FarnazH I'm merging this in a quicky because the full effect is only visible in other PRs after this gets merged. It is a bit of a learning curve...

PR #212, opened 2020-09-02
Document unit conventions

tovrstra:
I'm documenting the unit conventions because this was not entirely obvious. See #198.


tovrstra:
Thanks for taking a look. I'll merge.

PR #214, opened 2020-09-03
Turn all numedtuples into attrs classes

### LINKED ISSUE #201: Document required `attrs` package

tovrstra:
Fixes #201

Related to #138 and #157 (which were earlier attempts)

This PR includes:
- Attribute validation (to large extent, not every detail)
- attrutil module to facilitate validation of array attributes
- Documentation of how attrs is used in IOData
- Bug fix in CP2K loader, related to https://github.com/theochem/gbasis/issues/78
- Minor fixes elsewhere to satisfy attribute validators

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -338,7 +338,7 @@ validation, which eliminates potentially silly bugs.
 (See ``iodata/attrutils.py`` and the usage of ``validate_shape`` in all those
 classes.)
 
-The following ``attrs`` functions could be convenient with working with these
+The following ``attrs`` functions could be convenient when working with these
 classes:
 
 - The data can be turned into plain Python data types with the ``attr.asdict``
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -327,7 +327,78 @@ to avoid duplicate efforts.
     results in minor corrections at worst. We'll do our best to avoid larger
     problems in step 1.
 
+
+Notes on attrs
+--------------
+
+IOData uses the `attrs`_ library, not to be confused with the `attr`_ library,
+for classes representing data loaded from files: ``IOData``, ``MolecularBasis``,
+``Shell``, ``MolecularOrbitals`` and ``Cube``. This enables basic attribute
+validation, which eliminates potentially silly bugs.
+(See ``iodata/attrutils.py`` and the usage of ``validate_shape`` in all those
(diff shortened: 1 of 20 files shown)


tovrstra:
The previous two related PRs contain many outdated and redundant comments, which is the reason for making a ne PR. This is the cleaned-up version, which should be easier to review.

CONTRIBUTING.rst
```diff
@@ -327,7 +327,78 @@ to avoid duplicate efforts.
...
+--------------
+
+IOData uses the `attrs`_ library, not to be confused with the `attr`_ library,
+for classes representing data loaded from files: ``IOData``, ``MolecularBasis``,
+``Shell``, ``MolecularOrbitals`` and ``Cube``. This enables basic attribute
+validation, which eliminates potentially silly bugs.
+(See ``iodata/attrutils.py`` and the usage of ``validate_shape`` in all those
+classes.)
+
+The following ``attrs`` functions could be convenient with working with these
```
evohringer, line 341: > functions could be convenient with working with 

One with too much :-)

iodata/iodata.py
```diff
@@ -30,36 +31,9 @@
 __all__ = ['IOData']
 
 
-def convert_array_to(dtype):
```
evohringer, line 33: I like that these two functions are not part of iodata.py anymore and are now in atrs utils. This is cleaner and easier to follow.

evohringer (approved):
I´m not very familiar with the attrs class but it looks fine. Only a minor correction and a comment. Learnt  a lot reviewing :-)

CONTRIBUTING.rst
tovrstra: Thanks. Fixed.

iodata/iodata.py
tovrstra: Yep, they were indeed not in the right place, a bit too intimidating when one just wanted to see how the IOData class works. :)

tovrstra:
Several people have already reviewed this (including in former PRs), so I'm going to merge this.

PR #215, opened 2020-09-04
Fix exception and add unit test

tovrstra:
This just fixes a small snag.


iodata/test/test_extxyz.py
```diff
@@ -50,6 +50,17 @@ def test_load_fcc_extended():
     assert_allclose(mol.atcoords[31, 1], 5.56174271338 * angstrom)
 
 
+def test_load_mgo():
+    with path('iodata.test.data', 'mgo.xyz') as fn_xyz:
+        mol = load_one(str(fn_xyz), fmt='extxyz')
```
mcoolsce, line 55: This file is not yet present in the commit.

mcoolsce (commented):
Thanks for fixing the bug. It was not covered in the tests yet, but it should work now.

PR #216, opened 2020-09-07
Host docs on github

tovrstra:
This change will cause the documentation on github.io to be rebuilt with every release. I've tested it manually without creating a real release. See https://theochem.github.io/iodata/

### DIFF
--- a/.roberto.yaml
+++ b/.roberto.yaml
@@ -12,6 +12,7 @@ project:
         - pytest
         - upload-codecov
         - build-sphinx-doc
+        - upload-docs-gh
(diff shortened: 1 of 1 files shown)


FarnazH (approved):
This looks great to me! Is there anything else that needed to be added @tovrstra?

tovrstra:
@FarnazH This is all. Most of the work is done by Roberto, which was code previously. See https://github.com/theochem/roberto/blob/master/roberto/tasks.py#L366 (It is a fairly tedious procedure.)

PR #217, opened 2020-09-25
Add qcschema_molecule load_one and dump_one functions

### LINKED ISSUE #79: Add ordering from CCA to iodata.basis

wilhadams:
Broken off from closed PR #144.

This PR handles the molecule schema from QCSchema, which is the most mature of the 4 schema. 

I added the CCA conventions, so this closes #79


wilhadams:
The `load_one` function should be complete and ready for review. I'll try to get the `dump_one` up soon as well.

wilhadams:
The build is failing because of an E203 whitespace before ':', but in this case : is an operator so black will always revert it to that. I tried `#noqa` but that's for `flake8`, not `pycodestyle`.

PaulWAyers:
@wilhadams  is this *the* QC-Schema addition that needs to be reviewed?

wilhadams:
@PaulWAyers this is the QCSchema Molecule addition, it's been ready for a while, except for the issue with the linters I brought up above. Once it's reviewed I can rebase and squash in that last attempt to fix the linters, but the work here is complete.

tovrstra:
@wilhadams I'm currently running the tests locally because Travis-CI seems to be defunct at the moment. So far I've just addressed some linting issues but I will go through it in more detail soon.

(We have not used black on IOData, for historic reasons. It's a great tool but has some quirks too, like not producing PEP-8 compliant code and expanding code over too many lines.)

tovrstra:
@wilhadams I fixed things where possible. Could you check if these changes are good?

iodata/basis.py
```diff
@@ -367,3 +367,26 @@ def get_default_conventions():
 
 
 HORTON2_CONVENTIONS, PSI4_CONVENTIONS = get_default_conventions()
+
+# fmt:off
+CCA_CONVENTIONS = {
```
tovrstra, line 372: These are not yet used, which is fine, I guess? Just a few quick questions:
- Is the `fmt:off` related to black?
- Could you add a reference for these data? There may be some ambiguity. As far as I recall, CCA conventions only apply to Cartesian functions, not pure ones.

iodata/formats/json.py
```diff
@@ -0,0 +1,624 @@
...
+        keys and corresponding values as well.
+
+    """
+    # All Topology properties are found in the "molecule" key
+    molecule_dict = _parse_topology_keys(result, lit)
+
+    return molecule_dict
+
+
+def _parse_topology_keys(mol: dict, lit: LineIterator) -> dict:
```
tovrstra, line 186: What do you mean by "topology"? Is this QCSchema terminology?

tovrstra:
@wilhadams P.S. Please, do not rebase. It's not needed and causes more trouble than it is worth.

iodata/formats/json.py
wilhadams: Topology is essentially all the specifiers for a molecule. I'm using "topology", in line with how QCSchema does, because while the Molecule schema only uses these keys, the Input and Output schema also specify a molecule and will use this function to parse those keys.

iodata/basis.py
wilhadams: fmt:off is for black, I mostly used black throughout but not for this section, so that it didn't take 100 lines. 

I sourced the CCA conventions from the [libint wiki](https://github.com/evaleev/libint/wiki/using-modern-CPlusPlus-API#solid-harmonic-gaussians-ordering-and-normalization)

iodata/formats/json.py
tovrstra: Thanks for explaining!

wilhadams:
Thanks for the changes Toon, at this point I think this is good, is there anything missing?

iodata/basis.py
tovrstra: Thanks. I forgot about the Wiki.

tovrstra:
@wilhadams I've made one more change after checking out the details of the conventions. It turns out that CCA uses the same conventions as Psi4, which we already had implemented. I've renamed these to `CCA_CONVENTIONS` and added some info to the docstring. Unless you have comments on this last changes, we can merge. (I'll wait for your approval.)

wilhadams:
@tovrstra Looks good to me!

tovrstra:
Thanks!

PR #218, opened 2021-02-15
PDB format: Load & Dump Chain ID

FarnazH:
I was using `IOData` to load some `PDB` files, and needed the chain ID information but noticed that:

- `load_one` does not read chain ID
- `dump_one` uses `A` as a default for everything

So, I fixed these to parse chain id (if available), and when dumping a pdf file leave chain id empty if not available. I also polished the exiting code a bit to put the information parsed in the order they appear on a PDB ATOM line. 

Please feel free to let me know if anything needs to be undone.


FarnazH:
@tovrstra Travis failures are regarding `cython`, should the `environment.yml` be updated?

tovrstra:
Looks good. I'd only suggest including a test file with multiple chains. The following is a small PDB file with three chains: https://www.rcsb.org/structure/2BCW which seems good for testing.

lmacaya (dismissed):
@FarnazH thanks for the changes, considering that 'chain ID' column manipulation is very common. No modifications from my side.

tovrstra:
@FarnazH Can you take a quick look at the few changes I made? Thanks.

FarnazH:
Thanks for taking care of this @tovrstra. I started working on it when the PR was reviewed but it fell through the cracks and I didn't get to finalize and push it; I am sorry about that. 

I just added a test for dump_one with the multiple chain PDB file.

tovrstra:
Thanks. Don't get me started about my cracks. :)

tovrstra (approved):

PR #224, opened 2021-03-17
Add more flexible template substitute

tczorro:
Allow more fields to be added in custom template
Allow custom fields replacement in the default template


tovrstra:
@tczorro Thanks. I'm going to test locally and merge this despite not having tests for it because (i) it was an obvious typo and (ii) we should eventually think of a nicer solution for custom keywords, e.g. by placing them in `IOData.extra`.

tczorro:
@tovrstra Thank you for merging it. I was more thinking of putting miscellaneous contents for the input file such as `%chk` or other non-chemical related field.

tovrstra:
Thanks for explaining. I your case `kwargs` is indeed the most convenient option, so it should stay.

PR #231, opened 2021-03-22
Rename few variables for clarity

tovrstra:
@BradenDKelly I'm going to look into the overlap function to speed it up further. These are just a few clarifications I wanted to get out of the way first, which should be fairly harmless. Would you be able to quickly chem them? Thanks.

### DIFF
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -89,14 +89,14 @@ def compute_overlap(obasis: MolecularBasis, atcoords: np.ndarray) -> np.ndarray:
             # START of Cartesian coordinates. Shell types are positive
             result = np.zeros((len(scales[i0][0]), len(scales[i1][0])))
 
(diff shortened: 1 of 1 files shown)


BradenDKelly (approved):
Looks fine. At some point I think it would be good to modify the tolerance 1e-15 to be a default value, which a user can supplant, if they wish. That can be done later though.

On another note, I was working on a faster version, I broadcasted two more loops. 95% of the matrix values were correct, but there was a systematic error in a couple of values that I did not resolve. That too has been on my backburner. I will try to look at it as well. Memory is hazy, but I believe the speedup was fairly noticeable.

BradenDKelly (approved):

tovrstra:
Thanks for the quick check. I think I know how to eliminate a few explicit loops without too much trouble, so I'll give that a shot this week. It may also make things easier for you.

PR #236, opened 2021-03-24
Improve Input Writer docs & allow string template. 

### LINKED ISSUE #234: Add robust features to Gaussian input files
- [feature request] 

The current `write_input()` for Gaussian works well for typical jobs that require only having a theory, basis and simple task i.e., opt, energy etc. But a more complicated directive does not seem possible?

For instance if I want to make the following input file, is it possible?
```
%NProcShared=4
%mem=16GB
%chk=B3LYP_def2-qzvp_Be_13_O_14
#p nosymmetry B3LYP def2qzvp scf=(maxcycle=900,verytightlineq,xqc) integral=(grid=ultrafinegrid) pop=(cm5, hlygat, mbs, npa, esp) 

much fun, great quantum

-2 1
O 0.0 0.0 0.0
O 3.7 0.0 0.0
... (other atoms etc)
```
I mention this because I want to make these types of inputs. If this seems like a good thing to add, I can work on it and submit a PR.

Would a simple solution be to add a 'directive' attribute to 'IOData'?
`iodata.directive = "scf=(maxcycle=900,verytightlineq,xqc) integral=(grid=ultrafinegrid) pop=(cm5, hlygat, mbs, npa, esp) "`

similarly for specifying
```
%NProcShared=4
%mem=16GB
%chk=B3LYP_def2-qzvp_Be_13_O_14
```
these could be attributes or inside the extra dict?

wilhadams:
As discussed in #234 , we should allow a string input to the input writer. I haven't implemented that yet, I'm waiting for consensus on how we should differentiate between passing a string template and passing a filename, but I'm happy to implement it once we've decided which way to go about it. This PR will close #234 

In the meantime, I added a section to the "Supported Input Formats" page that shows the default template for each format. @tovrstra , feel free to directly fix any mistakes I made with that, but as far as I can tell it works correctly. 

I will also add a small section about the templates in the Getting Started page.


tovrstra:
My changes are really minor. I've used the new string formatting in Python instead of the `Template` class, which is something we might also consider for the input generation itself too. The `.format` method is much more powerful. Anyway, one step at a time.

I'm fine with already merging this as it is.

tovrstra (approved):

wilhadams:
Thanks for the changes @tovrstra 

I'll make the other changes to the writer now and push them soon.

wilhadams:
I added an example for using kwargs too, since Derrick added that in #224

tovrstra:
@wilhadams Thanks for this and for fixing the typos, very much appreciated. @BradenDKelly Can you check if this works for you?

BradenDKelly (approved):
Yep, it works. The **kwargs is a great feature and having an example using **kwargs in the docs is really nice.

tovrstra:
Thanks for the quick review!

PR #237, opened 2021-03-24
Fractional bond orders

### LINKED ISSUE #222: Add support for fractional bond orders
At the moment, bond orders are force to be integer because they are stored in the same array containing the atomic indices. This can be fixed easily with a structured data type: `np.dtype("int, int, float")`.

tovrstra:
Fixes #222 

@wilhadams This fixes the mismatch with QCSchema. I'm still putting all information in one array. This way we do not need to have attribute validators comparing the lengths of the arrays.

@evohringer I recall we initially discussed the bond attribute this with you. Would the current implementation work for your purposes?

### DIFF
--- a/iodata/formats/json.py
+++ b/iodata/formats/json.py
@@ -39,7 +39,7 @@ from warnings import warn
 import numpy as np
 
 from ..docstrings import document_dump_one, document_load_one
-from ..iodata import IOData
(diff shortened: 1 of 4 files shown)


tovrstra:
I tried to use this, but the API in this PR is rather inconvenient from the user perspective. Splitting up the bond information over two arrays would result in a much more intuitive API.

PR #238, opened 2021-03-25
Overlap integrals for a pair of basis sets

tovrstra:
@BradenDKelly It's becoming a daily routine. :) I did already try some other performance improvements, but further (significant) speedups without drastic code changes seem difficult to accomplish. For example, just vectorizing code inside the current bottleneck (`compute_overlap_gaussian_1d`) is not helpful because the arrays already too small at that level. Anyway, before doing anything more substantial, I'd figured it would be useful to add a feature to the overlap code, i.e. to compute overlaps between two basis sets, which are needed in #146 and #175. When trying to make things more efficient, we should not make it impossible to support this feature.

### DIFF
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -18,6 +18,8 @@
 # --
 """Module for computing overlap of atomic orbital basis functions."""
 
+from typing import Optional
+
 import attr
 import numpy as np
 from scipy.special import binom, factorial2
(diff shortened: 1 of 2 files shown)


BradenDKelly:
There does seem to be quite abit of `overlap` between additions... dun dun dun :)

Will look this over, unfortunately, probably on the weekend. Want to look it over carefully. Need to create and run some workflows involving the latest additions to `iodata` first though.

tovrstra:
No problem. Take your time.

BradenDKelly (approved):
I think this is good to go. Should not have said take my time, I totally lost track :S

tovrstra:
No problem. Thanks for reviewing!

PR #240, opened 2021-03-26
Fix import issue of setup.py

### LINKED ISSUE #239: `setup.py` imports IOData (indirectly), which may cause troubles
The `exec(f.read(), myglobals)  # pylint: disable=exec-used` triggers an import of IOData, which was not intended. It may break `setup.py` when some dependencies of IOData are not installed (or not up-to-date) before running `setup.py`. See https://github.com/QuantumElephant/databases/pull/7#pullrequestreview-621374680

tovrstra:
Fixes #239 More info follows later


tovrstra:
My previous fix for `setup.py` and the `__version__` issues did not seem to have work for everyone. This is an attempt to clean things up.

@BradenDKelly @shivupa @tczorro Would you be able to test if this branch is installable in various ways on your system? Part of the difficulty is that I cannot reproduce certain errors. Your help with testing would be very welcome to iron this out. There are a few methods to install a development version, all of which should work:

- Direct install from git. This is the simplest method, but it assigns a bogus version string, because pip has no infrastructure for assigning version strings from git tags (afaik). For this PR, the following should work, depending on the correctness of your `PATH` variable. If you have no virtual environment, you can always add the ``--user`` option:
  ```bash
  # First one also works with incorrect PATH
  python -m pip install git+https://github.com/tovrstra/iodata.git@fix-setup-import
  pip install git+https://github.com/tovrstra/iodata.git@fix-setup-import
  # Similar commands without venv or conda
  python -m pip install git+https://github.com/tovrstra/iodata.git@fix-setup-import --user
  pip install git+https://github.com/tovrstra/iodata.git@fix-setup-import --user
  ```
- Manual checkout of the branch, optionally setting the correct version string, followed by `setup.py` or `pip`:
  ```bash
  # Clone git repo with https or ssh:
  # The second one only works if you have ssh set up for Github
  git clone https://github.com/tovrstra/iodata.git
  git clone git@github.com:tovrstra/iodata.git
  # check out this feature branch of this PR
  cd iodata
  git checkout fix-setup-import
  # Optionally write the version string
  pip install roberto  # or any of the other ways of running pip, see above.
  rob write-version
  # Install can be done in several ways. --user is needed when you're not using venv or conda.
  ./setup.py install
  python setup.py install
  pip install .
  python -m pip install .
  ./setup.py install --user
  python setup.py install --user
  pip install . --user
  python -m pip install . --user
  ```

After this branch is merged, the official repo `https://github.com/theochem/iodata.git` and the master branch can be used instead of this feature branch. In case you installed Roberto, you can also run the entire CI workflow locally by executing `rob` in the root directory of the repository.

Once this is ironed out, I'll include similar instructions (for the master branch) in the documentation.

tczorro:
@tovrstra Thank you for the update. I tried on my Mac with Python3.7. Both `pip` and `python install` approach works properly with `0.0.0.post0` version number.

BradenDKelly:
Inside a virtual environment both the manual checkout and direct install from git worked. I have not tried installing into base yet.

shivupa:
Thanks! 

For me, directly installing from git gave the correct version number (using python 3.9.2 Linux no venv or conda). 

Also, the manual check out worked for me with and without `rob write-version`. I guess without `rob write-version`, it is essentially what the pip install from git is doing.

tovrstra:
I've added documentation for the installation from git. I'll leave this PR hanging for a few days and merge after the weekend if no new issues appear. In any case, if something seems wrong with the instructions at a later stage, feel free to open another PR.

@shivupa It is unlikely that pip calls `rob write-version`. Roberto is not widely adopted, which is an understatement, so pip will not care. It may happens that the version info from an earlier install is still lingering around in the source tree.

tovrstra:
The updated docs can be read more easily here: https://github.com/theochem/iodata/blob/73e7eb67978027498d67fae324fca4bfd8750985/doc/install.rst

PR #242, opened 2021-03-26
Use string `.format` method instead of `Template` module in `write_input`

### LINKED ISSUE #241: Use `.format` method instead of `template` module in `write_input`
The `.format` method is more powerful than the `template` module and would make it easier to add more features to the input writer without adding much more code. For example, the following is something only the `.format` method can do:

```python
print("access element from dict: {d[foo]}".format(d={"foo": "bar"}))
```

wilhadams:
This replaces the Template module with the format string method, as discussed in #241. 

I felt that it would be easier to unpack the `fields` dict into the `format` method, so that the template looks like
`#n {lot}/{obasis_name} {run_type}` and not `#n {fields[lot]}/{fields[obasis_name]} {fields[run_type]}`. I don't think that will cause any problems. 

I also added another example in the docs, to show that it's possible to read a template from file, in case someone is hoping to do that.

Closes #241

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -174,7 +174,8 @@ The run types can be any of the following: ``energy``, ``energy_force``,
 keywords when the file is written.
 
 It is possible to define a custom input file template to allow for specialized
-commands. This is done by passing a template string using the optional ``template`` keyword:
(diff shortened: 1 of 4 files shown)


tovrstra (approved):
Very nice. All is good.

PR #243, opened 2021-03-27
Adding additional bond type support when reading MOL2 files

fdroessler:
I noticed that the mol2 format reader only supports bond types 1, 2, 3 and am
However, there are a few more defined in the original specification. This PR will add the support for these to the code.

Similar to assigning "am" to 4 it will also just assign the other bond types to increasing integers.

### DIFF
--- a/iodata/formats/mol2.py
+++ b/iodata/formats/mol2.py
@@ -39,11 +39,12 @@ __all__ = []
 
 PATTERNS = ['*.mol2']
 
+
 # mol2 bond types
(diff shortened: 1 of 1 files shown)


tovrstra:
@fdroessler Thanks for making this PR. I didn't realize other bond types exited when reviewing the original code. There are a few minor issues in the changes, which I'll iron out.

tovrstra:
It seems I can't push to your PR, so I'll just open a separate one.

PR #248, opened 2021-03-29
Add support for bonds to SDF format

tovrstra:
This PR adds support for the bond information in the SDF format (both load and dump). I've also made some minor improvements to conform better to the official format.


iodata/formats/sdf.py
```diff
@@ -41,23 +47,36 @@
 PATTERNS = ['*.sdf']
 
 
-@document_load_one("SDF", ['atcoords', 'atnums', 'title'])
+SDF2BOND = {1: "1", 2: "2", 3: "3", 4: "ar", 5: "sd", 6: "sar", 7: "dar", 8: "un"}
+BOND2SDF = dict((val, key) for key, val in SDF2BOND.items())
```
evohringer, line 51: Could you maybe leave a comment why you had to define this dicitionary locally and not use the one in periodic. Just to have it commented. :-)

evohringer (approved):
It looks all great @tovrstra. Only a small suggestion to add a comment.

iodata/formats/sdf.py
tovrstra: We may be able to get rid of these two mappings, which may also be a good solution. The reason for having them is that MOL2 uses slightly different conventions, but we can bend those a little.

tovrstra: I've eliminated `SDF2BONDS` because it seemed better get rid of complicated things than having to explain them.

evohringer (approved):
This looks good to me!

tovrstra:
@evohringer Thanks for reviewing!

PR #249, opened 2021-03-30
Temporarily change install instructions in README

tovrstra:
I've commented out the install instructions for the stable release in replaced it by the install from git. This is a temporary change, until we have a first API-stable release.


FarnazH (approved):
Great & thanks @tovrstra. I just make a one-line commit to fix the commented section.

tovrstra:
@FarnazH Thanks!

PR #250, opened 2021-03-30
Add support for `CONECT` lines in PDB file (load and dump bonds)

tovrstra:
This adds support for the `CONECT` lines in pdb load and dump functions. I also split up the load_one functions over a few helpers to stop pylint from complaining about too long functions.


iodata/formats/pdb.py
```diff
@@ -62,66 +60,116 @@ def load_one(lit: LineIterator) -> dict:
...
-    title = "PDB file from IOData"
     while True:
         try:
             line = next(lit)
         except StopIteration:
             break
-        # If the PDB file has a title replace it.
+        # If the PDB file has a title, replace the default.
         if line.startswith("TITLE") or line.startswith("COMPND"):
             title = line[10:].rstrip()
```
FarnazH, line 130: If the PDB has a `TITLE` or `COMPND` section, they are usually split over multiple lines, and I think the current code only keeps the last line. So, shouldn't we use sth like: `title += line[10:].rstrip()`?

iodata/formats/pdb.py
```diff
@@ -62,66 +60,116 @@ def load_one(lit: LineIterator) -> dict:
...
+            restypes.append(restype)
+            chainids.append(chainid)
+            resnums.append(resnum)
+            atcoords.append(atcoord)
+            occupancies.append(occupancy)
+            bfactors.append(bfactor)
             molecule_found = True
+        if line.startswith("CONECT"):
+            for iatom0, iatom1 in _parse_pdb_conect_line(line):
+                bonds.append([iatom0, iatom1, 1])
```
FarnazH, line 145: Can you please clarify what `1` means here?  Is it meant to be bond-order? If so, what do you think of using `None` because it is not reported in PDB?

FarnazH (changes requested):
Thanks for the upgrade.

evohringer (changes requested):
I have only some minor issues. A test for the title over multiple lines would be nice I think. Everything else is great and improved it a lot!

iodata/formats/pdb.py
tovrstra: That would indeed be good. We may still include the newline characters, such that the title can be written out over multiple lines as well.

iodata/formats/pdb.py
tovrstra: I agree no bond type should be specified but with the current API, an integer must be provided, setting to `None` is not an option. The integers are "bond types" as defined in `iodata.periodic`. Setting it to `bond2num["un"]` (unknown) could be good temporary solution. My intention is to fix address this more generally with #222 (to introduce optional bond types and bond orders). With this PR and #248 I mainly wanted to get a better understanding of the different ways bonds are stored in various file formats. When addressing #222, this code will be updated.

iodata/formats/pdb.py
```diff
@@ -159,6 +209,22 @@ def dump_one(f: TextIO, data: IOData):
         out1 = f'{i+1:>5d} {attype:<4s} {restype:3s} {chain:1s}{resnum:>4d}    '
         out2 = f'{x:8.3f}{y:8.3f}{z:8.3f}{occ:6.2f}{b:6.2f}{n:>12s}'
         print("ATOM  " + out1 + out2, file=f)
+    # Prepare for CONECT lines.
+    connections = [[] for iatom in range(data.natom)]
+    if data.bonds is not None:
+        for iatom0, iatom1 in data.bonds[:, :2]:
```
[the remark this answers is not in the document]
tovrstra: Good point!

tovrstra:
@FarnazH All should be fixed. We may have a caching issue in the CI (for ubuntu-latest, 3.9), which is fixed in the master branch. There was no nice way to split `load_one` into small chunks (too many return values), so I've added pylint exceptions.

evohringer (approved):
Looks fine! Great!

tovrstra:
Thanks for reviewing!

Issue #191, opened 2020-08-23
General API for preparing IOData object before dumping


tovrstra:
**Motivation**

At the moment, it is mostly assumed that an `IOData` instance contains all the right attributes in the right form before it is passed on to a `dump_one` call. Some file formats (WNF, WFX and potentially also FCHK) modify the `IOData` object to become compatible with the format. Typical modifications include:

- WFX/WFN: converting basis to Cartesian functions (with transformation of the MO coeffs)
- WFX/WFN: decontracting the basis  (with transformation of the MO coeffs)
- FCHK: convert (natural) orbitals to density matrix, recontraction of basis sets.
- In general, possibly too far fetched, reverse-engineering contractions from WFN/WFX files.

This is problematic for several reasons:

- It makes `dump_one` functions long and complicated.
- Users may not be aware of the conversion taking place, which may result in loss of information.
- It may sometimes be of interest to disable conversions, e.g. when they are optional or when the user does not want any conversions (and prefers an exception to be raised instead). The latter is typical when dealing with conversions of large data sets, where data preservation is desirable and unintended loss of information due to conversion is not wanted.
- Some of the current conversions introduce redundant data, which results in inefficient use of storage.

See also:

- https://github.com/theochem/iodata/pull/164#discussion_r467966558
- https://github.com/theochem/iodata/pull/252#pullrequestreview-2099358284

**Proposal**

- Add an optional `prepare_dump` function to the fileformat modules. If present, it takes and `IOData` instance as argument, and returns a potentially modified one. The given `IOData` instance is not modified. An option `allow_changes=False` should be added, to allow disabling any conversion. If this flag is set to `False` and the file cannot be written without conversion, an exception is raised. If this flag is set to `True`, a warning is emitted when a conversion is applied.

- The `dump_one` and `dump_many` in the file formats functions call the new `prepare_dump` function before dumping.

- Add a `allow_changes=False` option to the `dump_one` and `dump_many` functions in the file formats modules. This is passed on to the `prepare_dump` function. The dump functions return the potentially modified `IOData` instance(s).

- Add a `allow_changes=False` option to the `dump_one` and `dump_many` functions in the module `iodata.api`. This is passed on to the dump functions of the selected file format. Also these dump functions return the potentially modified `IOData` instance(s).

- Factor out some of the reusable utility functions to modify the `IOData` object, e.g. manipulations of basis set and corresponding changes to MO coefficients.

[...]

PR #252, opened 2021-03-31
Improve treatment of occuptions numbers of restricted orbitals

### LINKED ISSUE #158: Bad alpha and beta occupation numbers
In the MolecularOrbitals class, there is an issue for alpha and beta occupations.

When the occupations are nearly integers (very close to 0, 1, and 2), then the present rule works.

When the occupations are fractional, it would be more sensible to choose alpha=beta occupations, which is correct for singlets and for the M_{s} = 0 (N_{alpha} = N_{beta}) state associated with doublets, triplets, etc..

I'd recommend detecting whether occupations are near-integer and, if not, using alpha=beta occupations instead (and printing a warning that the code is likely to be reliable except for restricted singlets).

tovrstra:
Fixes #158.

The following remained as it was:

- Single `MolecularOrbitals` class for all kinds of orbitals.
- Restricted orbitals with integer occupation numbers are assumed to represent a restricted open- or closed-shell wavefunctions. E.g. occupation numbers `[2, 2, 1, 0, 0]` would still represent a doublet state.

The following has changed:

- Restricted orbitals with fractional occupation numbers are assumed to be spin-paired or closed-shell, unless specified otherwise. E.g. occupation numbers `[2, 2, 1.8. 0.2, 0]` are assumed to be spin-paired.

- It is possible to deviate from the above heuristics be specifying `mo.occs_aminusb`. Just keep in mind that most file formats cannot handle this case. They essentially rely on the above heuristics. Writing a wavefunction file will normally fail when `occs_aminusb` is set.


tovrstra:
@FarnazH @PaulWAyers It should be ready for review now. The few remaining coverage misses are fairly harmless.

tovrstra:
This is one of the changes that should be done before 1.0.0 because it changes the API. I've brought this zombie-PR back to live by merging the main branch into this PR, so the PR is up-to-date with the recent tooling.
(Deepsource complains about complexity of functions, but this problem is not due to the current PR.)

@FarnazH @PaulWAyers It would be good to have your comments on this one before it gets merged.

iodata/orbitals.py
```diff
@@ -168,32 +202,115 @@ def spinpol(self) -> float:
...
     @property
     def occsb(self):
-        """Return beta occupation numbers."""
+        """Return beta occupation numbers.
+
+        Notes
+        -----
+        For restricted wavefunctions, in-place assignment to occsb will not
+        work. In this case, the array is derived from ``mo.occs`` and optionally
+        ``mo.occs_aminub``. To avoid that in-place assignment of occsb is
```
PaulWAyers, line 273: mo.occs_aminusb (typo)

PaulWAyers (commented):
I realizes that the `occs_aminusb` is the most general form, but in the end this still reduces to `occs_a` and `occs_b`, so I'm a bit confused why methods that support unrestricted natural orbitals (like `wfx`) can't support this. It seems to me that for most methods that support non-integer occupations and unrestricted orbitals, `occs_aminusb` is also supportable, since it fully specifies `occs_a` and `occs_b`. 

Probably I'm missing something silly. I looked at `wfx` and didn't see why it could be supported, however. I didn't dig deeply as I'm on deadline this morning.

tovrstra:
Thanks for the comments. I'll look up the reason for why some formats are not supporting `aminusb`, if any. (If there is a good reason, it should be explained better in the comments.)

tovrstra:
@PaulWAyers I took a closer look at the code to see how we can address your concern. In the `dump_one` functions, when `occs_aminusb` is set, we could convert the IOData instance from a restricted to unrestricted set of orbitals. After that conversion, everything can proceed as normal. The conversion is needed because WFX, WFN, MOLEKEL and MOLDEN formats do not support unrestricted occupation numbers with restricted orbitals. (FCHK does.)

There are a few downsides to such a conversion (at this stage):

- The resulting files contain a lot of redundant information. At the very least, there should be warning, so users are made aware of this.
- In general, we don't always do such conversion automatically, to avoid losing the original representation of the data upon writing a file. For this reason, I've created issue #191.

My suggestion would be to not add the conversion at this stage and defer it until there is a proper general solution in place as discussed in #191. (I just edited that issue for clarity.)

iodata/orbitals.py
tovrstra: Thanks! Both occurrences of this typo were fixed.

PaulWAyers (approved):
Sure, we can address this later. The issue #191 is pretty major (and rather stale) so hopefully we get around to it. It might be a good GSoC project one year.

tovrstra:
Thanks for the review! I'll merge this one and add a comment on the aminusb situation to #191. We have a good number of stale issues, but I love cooking with stale bread. :)

PR #267, opened 2021-05-22
Fix missing `M END` tag for SDF format, close #266

### LINKED ISSUE #266: Fix `dump_one` and `dump_many` for SDF file format
The current implementation of `iodata` has some problems with saving SDF files.

- [x] No **M END** tag and therefore not readable by other packages.
- [ ] No *connectivity information* dumped and therefore, it is regarded as a set of atoms when loading with other packages.

I don't know a quick fix to it yet, but it would nice to have SDF dump functionality.

fwmeng88:
Because we are missing **M END** tag when dumping a molecule, the dumped molecule can not be recognized by molecular packages, such as `OpenBabel` or `RDKit`. This PR fixes this problem and should close #266.


tovrstra (approved):
Thanks. Looks good.

Issue #273, opened 2021-06-14
CFOUR molden files


shivupa:
Hi, 
We have noticed the MOs printed in the CFOUR MOLDEN output are not normalized as expected. We were interested in standardizing this.

We summarize the quirks below (relevant files are at this [gist](https://gist.github.com/shivupa/0905319e6492f297fbf17766bdaa3813)):

- CFOUR outputs "[Molden Format]white space" rather than [Molden Format]". Needs .strip()
- CFOUR also puts "[Molden Format]" in the file twice.
- CFOUR converts the AO basis back to the cartesian basis (regardless of if it was run in the spherical basis)

We have run zero electron calculations with d,f,g,h functions individually to assist with the normalization and ordering. Could you recommend a path forward to get the properly normalized MOLDENs out? We are willing to help as much as we can since this is critical for our research.

Thanks,
Shiv Upadhyay (@shivupa) and Amanda Dumi (@amandadumi)

PaulWAyers:
Thanks for pointing this out. @tovrstra is probably the best person to respond in detail. As you've surmised, molden files are darn tricky. Just off hand, however, I don't think it will be that difficult to support CFOUR molden files too.

tovrstra:
@shivupa Thanks for bringing this up. We'd love to update our code so it can handle CFOUR files. How familiar are you with changing the source code and making a pull request? We typically review PRs and make suggestions for code improvements until it all looks good. A good start is to create a PR with unit tests with CFOUR files, which just show that the current code fails. Then you can add commits to fix the problems. A technically detailed outline can be found here: https://github.com/theochem/iodata/blob/master/CONTRIBUTING.rst

I took a quick look at the molden file in the gist and this one is certainly useful for unit testing, i.e. it is not too big. It will take a bit of reverse engineering to find out the non-standard conventions that CFOUR Molden files use. By calculating and printing out the overlap matrix with IOData of the MO's for a small calculation, like you provide, you can detect where the issues are. This should obviously be an identity matrix, but this fails when there is a mismatch with the conventions. An example of such a test can found here:

https://github.com/theochem/iodata/blob/1de466118c3fbc0017a69f00fe896cb9134d8bc1/iodata/test/test_molden.py#L263-L281

The function `check_orthonormal` does that test. Also the `@pytest.mark.parametrize` can be convenient to write a single test for multiple files.

The function with all the fixes for buggy molden files can be found here:

[...]

PR #276, opened 2021-07-11
Cfour molden

amandadumi:
We expanded the Molden support to include those generated from CFOUR as discussed in #273 . There is support for orbitals up to g angular momenta . The unit test ensures the MOs are orthonormal for a set of zero electron calculations, each ran with one basis function. The set of Molden files span angular momenta from s to g.


tovrstra:
Thanks for the PR! (I wasn't aware I had to approve running the tests, so these just started.) I'll take a closer look soon.

tovrstra:
The small changes I made should get us a bit further with the CI, but I still need to review more carefully.

tovrstra:
@amandadumi One test is failing, seems to be due to a missing file. Could you add a commit with the missing file? Thanks! Details can be found here: https://github.com/theochem/iodata/pull/276/checks?check_run_id=3043813103#step:6:1454

amandadumi:
Yes, of course! I changed the file names at the last second to be more clear and forgot to update the unit test file. I will fix this shortly, thank you for such quick feedback.

tovrstra (approved):
Thanks! merging...

PR #279, opened 2021-08-27
Fix potential typo in JSON loader

tovrstra:
@wilhadams Could you check if this was indeed a typo? (It's just on one line.) If Yes, I'll see If I can easily add a test for it.


wilhadams (commented):
Hi Toon, thanks for checking this, but the Input schema (to the extent that it actually exists within MolSSI's implementation) is still at this point version 1 to my best knowledge. 

It's hard to know for sure, QCElemental only has the current version, the GitHub repo v2 says that the version is "1.dev", and the website doesn't have any versioning info nor an up-to-date example.

tovrstra:
Thanks for checking and I agree it is a bit confusing. :) I'll close.

PR #280, opened 2021-09-12
Add QCSchema Basis

wilhadams:
This adds the QCSchema basis load_one and dump_one. I wouldn't ever use this schema over the Basis Set Exchange's basis schema, which is just slightly incompatible with this schema for unknown reasons. In future it would probably be better to use the BSE python module for converting basis set data.


tovrstra:
One test is failing: https://github.com/theochem/iodata/pull/280/checks?check_run_id=3584491841#step:6:1564

iodata/formats/json.py
```diff
@@ -976,55 +977,132 @@ def _find_passthrough_dict(result: dict, keys: set) -> dict:
...
+        center_shells[center] = list()
+        # QCElemental example omits ecp_electrons for cases with default value (0)
+        if "electron_shells" not in center_data[center]:
+            raise FileFormatError(
+                "{}: Basis center {} requires `electron_shells` key".format(lit.filename, center)
+            )
+        if "ecp_electrons" in center_data[center]:
+            ecp_electrons = center["ecp_electrons"]
+        else:
+            ecp_electrons = 0
```
tovrstra, line 1059: `ecp_electrons` is not used.

iodata/formats/json.py
```diff
@@ -976,55 +977,132 @@ def _find_passthrough_dict(result: dict, keys: set) -> dict:
...
+            # Load shell data
+            if shell["harmonic_type"] not in {"cartesian", "spherical"}:
+                raise FileFormatError(
+                    "{}: `harmonic_type` must be `cartesian` or `spherical`".format(
+                        lit.filename,
+                    )
+                )
+            angmoms = shell["angular_momentum"]
+            exps = np.array(shell["exponents"])
+            coeffs = np.array([[x for x in segment] for segment in shell["coefficients"]])
```
tovrstra, line 1079: May be simplified?
```suggestion
            coeffs = np.array(shell["coefficients"])
```

tovrstra:
@wilhadams Thanks for the PR. Please let me know if you'd like me to review it a more in depth.

PR #285, opened 2021-10-20
Add MO-basis RDMs fields

msricher:
Hi, @PaulWAyers @FarnazH, this is my modification to allow IOData to handle MO-basis RDMs. Thoughts?
I will then add a wrapper to PyCI/PySCF and put it in IOData/Scripts, so that we can use IOData with PyCI to run HCI computations.


tovrstra (approved):
Looks good. Thanks for this change! My apologies for the long delay.

PR #286, opened 2021-11-20
user-set threshold

amandadumi:
Introduces the option to allow the user to set the threshold for normalization if desired. If no threshold is provided, the default as previously defined is used.

A note: The threshold argument was an option for the `_is_normalized_properly` did have a threshold argument, but wasn't easily accessible by the user it seemed.


amandadumi:
ah, i see this function is called in other places. I'll look at addressing these over the weekend

tovrstra (approved):

tovrstra:
@FarnazH @PaulWAyers Some points were improved, and I'm fine with it now. I'll leave some time for additional comments until May 15 before merging.

Issue #269, opened 2021-06-03
Support *.mol files


PaulWAyers:
@vcanogil suggested that we should also support a *.mol file.

Links to the detailed file specification are below, but this should be *very* easy I think, because we already support a *.sdf file.

- https://chem.libretexts.org/Courses/University_of_Arkansas_Little_Rock/ChemInformatics_(2017)%3A_Chem_4399_5399/2.2%3A_Chemical_Representations_on_Computer%3A_Part_II/2.2.2%3A_Anatomy_of_a_MOL_file

- https://docs.chemaxon.com/display/docs/mdl-molfiles-rgfiles-sdfiles-rxnfiles-rdfiles-formats.md

- https://en.wikipedia.org/wiki/Chemical_table_file

nwoodweb:
Hi @PaulWAyers 
I am currently working on this issue

While I do have a functioning 'load_one' at this moment, I am working to see how I can go about parsing charges from the atom block

nwoodweb:
@FarnazH 

I wanted to inform you that I had passed the Roberto CI check and I will be making a pull request. 
While I have implemented an `iodata/formats/mol.py` file, I have not implemented a test file that other format files had according to Roberto and I was wanting to know if I should also implement this. 
![robertoMol-cropped](https://user-images.githubusercontent.com/60708568/171890266-e89f3c3f-ecd8-4b27-be3d-28fc1b6a4478.png)

PR #290, opened 2022-06-03
Added .mol functionality

nwoodweb:
The file `iodata/formats/mol.py` implements V2000 *.mol compatibility, based heavily off of currently implemented *.sdf compatibility

- I had used load_one on a phenol.mol file I had made
- I had passed Roberto CI check


FarnazH:
@nwoodweb thanks for your contributions. As shared in issue #269, having a test for every line of code is part of our contribution guidelines, so we cannot review your PR until suitable tests are added.

nwoodweb:
@FarnazH 
Ok thankyou for clarifying.
I will add a testing script in `iodata/test` as well as at least 2 example files  and a V2000 unit test  in `iodata/test/data`

nwoodweb:
I accidentally pushed to master, which is part of the pull request, then pushed my unit tests to `new-feature` so I had simply merged `new-feature` with my forks `master`

I added unit test file `iodata/tests/test_mol.py` as well as various example mol files

tovrstra:
Thanks for the contribution and my apologies for the delay. The code is almost identical to that for the SDF format. When looking for the differences, I could only find one thing related to the end marker for a single molecule:

- For SDF this is
  ```
  M  END
  $$$$
  ```
- For SDF this is
  ```
  M  END
  ```

Are there any other differences? If not, I'd suggest extending the SDF code to also handle the `.mol` case correctly. This way, we can avoid duplicated code.

nwoodweb:
Hi @tovrstra 

`.sdf` and `.mol` are for the most part identical besides the termination string at the end. In hindsight it might be better to use branch statements to differentiate `M END` and `M END $$$$`.

However, I do remember being confused about V2000 and V3000 formatting, as well as the bond block, so I will review those before implementing your recommendation.

PR #296, opened 2023-05-05
Fix DeprecationWarning: path is deprecated.

FarnazH:
DeprecationWarning: path is deprecated. Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.

PR #297, opened 2023-05-06
Depend on importlib_resources for py37 for as_file api

tovrstra:
This is a fix for py37 related to the recently merged #296

See https://github.com/theochem/iodata/actions/runs/4900272755/jobs/8750814221#step:6:933

I've locally reproduced the bug and tested the fix with py37.

It is a minor change, so I'll YOLO-merge this in the absence of comments after May 10.


PaulWAyers (approved):
Fine by me.

PR #301, opened 2023-11-18
Add workaround for scipy.special.factorial2 changes

FarnazH:
This PR adds a `factorial2` function that wraps `scipy.special.factorial2` to return 1.0 when the input is not positive. This is a temporary addition, while we wait for Scipy's update. To learn more, see https://github.com/scipy/scipy/issues/18409. This is a better solution than limiting Scipy version that works with IOData, as one might need to use scipy `1.11.x` in their environment for other purposes. A similar change was made in other QC-Devs packages; see https://github.com/theochem/gbasis/pull/138.

### DIFF
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -22,13 +22,31 @@ from typing import Optional
 
 import attr
 import numpy as np
-from scipy.special import binom, factorial2
+import scipy.special
 
 from .overlap_cartpure import tfs
 from .basis import convert_conventions, iter_cart_alphabet, MolecularBasis
 from .basis import HORTON2_CONVENTIONS as OVERLAP_CONVENTIONS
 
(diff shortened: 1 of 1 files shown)


PaulWAyers (commented):
Shouldn't we return 1 for factorial2(-1) and 0 for factorial2(x) where x = -2,-3,....? I think that is the "old" `scipy` behavior.

Should we check that n is an integer?

FarnazH:
@PaulWAyers, thanks for your feedback.
- We don't need to check that `n` is an integer, because this is just a wrapper function, and `scipy.special.factorial2` does that check and prints a proper error message: `ValueError: factorial2 does not support non-integral scalar arguments`. 
- Following your suggestion, I updated the function to return 0.0 for `n <= -2`; see b3b7377. As `scipy` does not treat odd and even negative integers differently, I left it intact.

Issue #308, opened 2024-02-08
Fix Factorial2


Ali-Tehrani:
The most recent merged [pull-request](https://github.com/theochem/iodata/pull/301) causes 165 tests to fail in Github [Actions](https://github.com/theochem/iodata/actions/runs/7161627372/job/19497490702#step:6:9878) for various python systems.  The following error is obtained for the vast majority of them:
```python
     def factorial2(n, exact=False):
        """Wrap scipy.special.factorial2 to return 1.0 when the input is -1.
    
        This is a temporary workaround while we wait for Scipy's update.
        To learn more, see https://github.com/scipy/scipy/issues/18409.
    
        Parameters
        ----------
        n : int or np.ndarray
            Values to calculate n!! for. If n={0, -1}, the return value is 1.
            For n < -1, the return value is 0.
        """
        # Scipy  1.11.x returns an integer when n is an integer, but 1.10.x returns an array,
        # so np.array(n) is passed to make sure the output is always an array.
        out = scipy.special.factorial2(np.array(n), exact=exact)
>       out[out <= 0] = 1.0
E       TypeError: 'int' object does not support item assignment
```

Further the following [line](https://github.com/theochem/iodata/blame/6b9c6e0e94a0425593217954f3913f4545a5605d/iodata/overlap.py#L240) does not match [scipy.factorial2](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial2.html) API.  Looking at the earlier release of this package, we want `exact=True` in scipy.factorial2.  When setting `exact=True`, then scipy.factorial2 only accepts integers, and so we will need to run it over each individual element of the array.

The solution would be too in `factorial2` function:
1. Keep `out = scipy.special.factorial2(np.array(n), exact=exact)` when exact=False
2. When exact=True, and 
   i. Add if n is not `int`, then convert it to array `n = np.array([n])`) 
   ii.  compute `scipy.special.factorial2(x, exact)` on each individual element `x` of the array `n`.
   
Thank you to @LudoRNLT and @D-TheProgrammer, if you got the time to look into this.

PaulWAyers:
@Ali-Tehrani is this also going to be an issue in [`GBasis`](https://github.com/theochem/gbasis/blob/9f6afe7ffc56b9a4e7f8a4a74d0fa5f01197ebca/gbasis/utils.py#L7)?

Ali-Tehrani:
[...]

PR #309, opened 2024-03-07
Update overlap.py to fix_factorial2

LudoRNLT:
Capture d'écran 2024-03-07 104652
Capture d'écran 2024-03-07 104853

@D-TheProgrammer and I, think that we solved the factorial error. Let us know if we have to do this but in a different way or use another syntax.

### DIFF
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -48,7 +48,7 @@ def factorial2(n, exact=False):
     if exact == False : #Keep when exact=False
         out = scipy.special.factorial2(np.array(n), exact=exact)
     else :
-        if n is not int:
+        if not isinstance(n, int):
             n = np.array([n]) # convert n to a np.array
             for x in range(len(n)):
                 x = n[x]
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -45,9 +45,14 @@ def factorial2(n, exact=False):
     """
     # Scipy  1.11.x returns an integer when n is an integer, but 1.10.x returns an array,
     # so np.array(n) is passed to make sure the output is always an array.
-    out = scipy.special.factorial2(np.array(n), exact=exact)
-    out[out <= 0] = 1.0
-    out[out <= -2] = 0.0
+    if exact == False : #Keep when exact=False
+        out = scipy.special.factorial2(np.array(n), exact=exact)
+    else :
+        if n is not int:
+            n = np.array([n]) # convert n to a np.array
+            for x in range(len(n)):
+                x = n[x]
+                out = scipy.special.factorial2(x, exact) #compute on each individual element x of the array n
     return out
 
 
--- a/iodata/overlap.py
+++ b/iodata/overlap.py
@@ -45,7 +45,7 @@ def factorial2(n, exact=False):
     """
     # Scipy  1.11.x returns an integer when n is an integer, but 1.10.x returns an array,
     # so np.array(n) is passed to make sure the output is always an array.
-    if exact == False : #Keep when exact=False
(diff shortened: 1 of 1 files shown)


PaulWAyers (approved):
This looks good to me. I've tagged @FarnazH as she did more work on this, so I'd wait a couple days for her to glance at it before merging.

D-TheProgrammer:
Hi @PaulWAyers  and @FarnazH , @LudoRNLT and I wanted to know if you have any updates regarding our fix and if it still suits you for the merge.

PaulWAyers:
I think @FarnazH is busy. Perhaps @tovrstra or @marco-2023 has time to do a quick review.

tovrstra:
@D-TheProgrammer: can you explain the error you were getting and that was fixed with this change? I just ran all unit tests without this change, and they pass (except for the problem fixed in #310).

iodata/overlap.py
```diff
@@ -45,9 +45,14 @@ def factorial2(n, exact=False):
     """
     # Scipy  1.11.x returns an integer when n is an integer, but 1.10.x returns an array,
     # so np.array(n) is passed to make sure the output is always an array.
-    out = scipy.special.factorial2(np.array(n), exact=exact)
-    out[out <= 0] = 1.0
-    out[out <= -2] = 0.0
+    if exact == False : #Keep when exact=False
+        out = scipy.special.factorial2(np.array(n), exact=exact)
+    else :
+        if n is not int:
```
tovrstra, line 51: Please fix the type checking:
```suggestion
        if not isinstance(n, int):
```

iodata/overlap.py
```diff
@@ -45,9 +45,14 @@ def factorial2(n, exact=False):
...
+            n = np.array([n]) # convert n to a np.array
+            for x in range(len(n)):
+                x = n[x]
+                out = scipy.special.factorial2(x, exact) #compute on each individual element x of the array n
```
tovrstra, line 55: Shouldn't this loop be dedented by four spaces?

iodata/overlap.py
```diff
@@ -45,9 +45,14 @@ def factorial2(n, exact=False):
     """
     # Scipy  1.11.x returns an integer when n is an integer, but 1.10.x returns an array,
     # so np.array(n) is passed to make sure the output is always an array.
-    out = scipy.special.factorial2(np.array(n), exact=exact)
-    out[out <= 0] = 1.0
-    out[out <= -2] = 0.0
+    if exact == False : #Keep when exact=False
```
tovrstra, line 48:
```suggestion
    if not exact:
```

tovrstra (changes requested):
I'm sorry for my initial confusion. I'd like to request a few improvements:

- Major: Please add unit tests for the cases you're fixing. The issue reported in #308 is not covered by our unit tests at the moment. Unit tests for the following cases (once with `exact=True` and once with `exact=False`) would be useful:
    - integer arguments `0`, `1`, `2` and `3`
    - float arguments `0.0`, `1.0`, `2.0` and `3.0`
    - integer array argument `np.array([0, 1, 2, 3])`
    - float array argument `np.array([0.0, 1.0, 2.0, 3.0])`
- Minor: Please avoid comments at the end of a line. Put them on the line before instead. Also, don't add comments with little added value, such as translating a line of Python to a human-readable sentence. Comments are more useful when they explain the intent of the code, typically of a block consisting of multiple lines, which is different from a literal human translation of the code.

Thanks!

tovrstra:
I'm closing this one because a more recent pull request was made: #319

PR #310, opened 2024-05-03
Fix Python 3.12 compatibility

tovrstra:
The `distutils` package is deprecated and no longer available in Python 3.12. We only used the `strtobool` function from `distutils`. This PR adds a `strtobool` replacement to `iodata.utils` and changes the imports from `distutils` by imports from our own `utils.py`.

I've also added `.envrc` to `.gitignore` because it is a local file I'm often using to manage different software envs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,9 @@ celerybeat-schedule
 # dotenv
 .env
 
+# direnv
+.envrc
+
 # virtualenv
 .venv
 venv/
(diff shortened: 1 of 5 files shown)


PaulWAyers (approved):
Looks good to me.

evohringer (approved):
Everything looks fine. Thanks Toon for taking care.

tovrstra:
Thanks for reviewing!

PR #312, opened 2024-05-03
Initiate switch to more modern GitHub Actions

tovrstra:
This builds on the changes in #310, which will be merged before this one.

Main changes:
- Replace `setup.py` by `pyproject.toml`
- Replace `.github/workflows/ci.yaml` by `.github/workflows/pytest.yaml`
- Remove unused config files

There are still related changes needed to get more linting back to work, but I'd like to defer this to later pull requests to limit the number of changes in one pull request. Things to be done include, probably one PR each:
- Include `ruff format` in the pre-commit config (and exclude some sections in the unit tests that would become poorly formatted) + lots of changes due to the automatic formatting.
- Include ruff in the pre-commit config + some cleanups due to ruff.
- Add a workflow for PyPI releases.

After this PR, the `master` branch will be renamed to the `main` branch. After the planned PRs listed here, I will update the other PRs.


tovrstra:
Some more background can be found in the discussion of pull request #263.

PaulWAyers:
Hi @tovrstra . I trust your judgement on this. As you note, the need to update our GitHub actions is essentially the conclusion from the *long* discussion in #263. 

As we agreed in February, we want to harmonize workflows so that all the QC-Devs packages have the same workflow; it is too hard to maintain if `iodata` is different from everything else. So modernizing GitHub Actions for `iodata` is, in practice, potentially a large task, as it requires doing the same for ~10 core packages (and also potentially updating the [contributing guidelines](https://github.com/theochem/guidelines/blob/main/contributing.md) and other organization-level documentation so that users can effectively use the new workflow). (I don't see any point in rolling out the changes to stale repositories or those we are keeping only for archival purposes.) I, unfortunately, am not going to be very helpful for this (sort of) task. 

@tovrstra can you coordinate the rollout of these changes across repositories, updating the contributing guidelines (if required) and documenting the workflow? (I leave it to your judgement as to whether we should first make sure everything works for `iodata` before rolling out changes organization-wide, or whether we should roll them out piecemeal to minimize disruption.) It may also be that we need to set up a "template repository" for new projects.

PaulWAyers (dismissed):
OK by me.

tovrstra:
Thanks for the comments. I will open a new issue to maintain an overview of the pending and completed updates, of which this is the first one.

Issue #313, opened 2024-05-14
Switch to more modern project structure with GitHub Actions


tovrstra:
This issue replaces some older ones and is used to keep track of several pull requests to reorganize the repository.

I will edit this TODO list to keep it up to date with merged pull requests and suggestions made below:

- [x] Replace `setup.py` by `pyproject.toml`. Fixed in #312. 
    - For some background, see: https://peps.python.org/pep-0518/
    - Version numbers are derived from Git tags with [setuptools-scm](https://pypi.org/project/setuptools-scm/).
    - [x] Also import `__version_tuple__` from `._version` See #315
- [x] Run unit tests and oldest and most recent Python version we'd like to support (3.9 ... 3.12). Fixed in #312
- [x] Rename `master` branch to `main`.
- [x] Add a branch protection rule that forces contributing branches to be made from forks. This prevents the original repository from being overloaded with branches, as is currently the case: https://github.com/theochem/iodata/settings/rules/772871
- [x] Add basic branch protection to the main branch: no force push, no delete. See https://github.com/theochem/iodata/settings/rules/772869
- [x] Add `ruff format` to pre-commit config. See https://docs.astral.sh/ruff/formatter/ (Almost identical to black, faster, will be integrated into `ruff` in the future) See #314 
- [x] Add `ruff` to pre-commit config. (Instead of all other linting tools) See https://docs.astral.sh/ruff/ See #315
- [x] Enable pull request review automation tools: deepsource, sonarcloud, sourcery-ai
- [x] Remove all `importlib_resources` try-except blocks. See #323
- [x] Caputre all warnings in unit tests. This is needed because:
    - Warnings often hint at problems that will become real errors in future version of dependencies. Hence, warnings should result in failed tests, so we certainly fix them before they become a real problem. See #329
    - Warnings can also be the consequence of antipatterns. See #327
    - Some warnings are generated in IOData as user feedback, which should be checked in the tests. See #326 and #328
    - Make pytest exit with an error when there are warnings. See #331
- [x] Update from attr to attrs. See #318 
- [x] Add GitHub actions for PyPI releases. See #317 
    - [x] Initiate a changelog file using the following format: https://keepachangelog.com/en/1.1.0/ See #322
    - [x] Yank `iodata`  releases on PyPI, so there is only the `qc-iodata` package on PyPI
    - [x] Update PyPI classifiers. See #324 
    - [x] Make new alpha release and document how to install it (README and sphinx doxs) See #325
[...]

PR #314, opened 2024-05-14
Ruff format

tovrstra:
This pull request takes care of the autoformatting mentioned in #313. I'm using [Ruff format](https://docs.astral.sh/ruff/formatter/) instead of [Black](https://black.readthedocs.io/en/stable/index.html). Both are very similar, but Ruff format is faster. Because I'd like to use Ruff also for linting (which is insanely fast and extensive) in the following pull request, it seems wise to use the same tool for as many tasks as possible.

This is a big change, but all of it is just formatting. I've excluded a few blocks from the autoformatting. See (`# fmt: off` and `fmt: on`) in 11c6203.

I've also upgraded a few other pre-commit tools to more recent versions, but this had any visible effect.

Tests pass on my machine.


tovrstra:
I'm planning to YOLO-merge this Wednesday, May 22, unless reviewed and approved earlier.

PaulWAyers (commented):
I'm OK with the switch to `ruff`. I've had both `ruff` and `black` locally for a while and can't see much difference. `ruff` seems more integrated (closer to one-stop shopping). Disadvantages I see from changing are:
- change is hard.
- ruff is developed by a for-profit company (unlike black)
- even the latest version of ruff (0.4) is still not a version 1.0-type stable release.

It would be annoying to change to ruff and then have to change back because the company died or started charging money. But that is probably a minor concern.

tovrstra:
I agree that there is not much information on how the company behind Ruff will survive in the long run, but given the popularity of the Ruff, it will probably outlive the company and remain available as open source in some form. Upgrading is quite smooth. I've gone through upgrades from pre-0.1 to 0.4 and barely noticed. The usage is more stable than the version numbering suggests.

PR #315, opened 2024-05-22
Ruff

tovrstra:
This is again a big change, cleaning up many small things with Ruff automatically. This removes a lot of cruft. In short, I'm using Ruff for this instead of many other linters because it is fast and it has the most comprehensive rule set implemented, many of which have automatic fixes. The complete list can be found here:  https://docs.astral.sh/ruff/rules/

Only the first commit (Basic Ruff configuration) required some decisions that are worth reviewing. All remaining commits are rather mindless.

I'm planning to YOLO-merge this Wednesday, May 29, unless reviewed and approved earlier.

For the bigger picture, see #313


tovrstra:
I've added fixes for NPY already, quite simple. I've removed FURB because it is apparently in preview.

PR #316, opened 2024-05-24
New version overlap.py to fix_factorial2 #309  with unittest named test_factorial2

D-TheProgrammer:
@LudoRNLT and I think that we have solved the fix for issue #309 described in issue #308 for this pull request. As suggested, we have added a unit test, and our code is passing both the `test_factorial2` and `test_overlap` unit tests. Please let us know if this is good and if it will be merged. Thank you.
Capture d’écran du 2024-05-24 21-50-39

Capture d’écran du 2024-05-24 21-54-10

### DIFF
--- a/iodata/test/test_factorial2.py
+++ b/iodata/test/test_factorial2.py
@@ -4,8 +4,8 @@ from .. import overlap
 
 factorial2 = overlap.factorial2
 
-class TestFactorial2(unittest.TestCase):
 
+class TestFactorial2(unittest.TestCase):
     def test_integer_arguments(self):
         self.assertEqual(factorial2(0, exact=True), 1)
         self.assertEqual(factorial2(1, exact=True), 1)
@@ -22,14 +22,12 @@ class TestFactorial2(unittest.TestCase):
 
     def test_integer_array_argument(self):
         np.testing.assert_array_equal(
-            factorial2(np.array([0, 1, 2, 3]), exact=True),
-            np.array([1, 1, 2, 3])
+            factorial2(np.array([0, 1, 2, 3]), exact=True), np.array([1, 1, 2, 3])
         )
 
     def test_float_array_argument(self):
         np.testing.assert_array_almost_equal(
-            factorial2(np.array([0.0, 1.0, 2.0, 3.0]), exact=False),
-            np.array([1.0, 1.0, 2.0, 3.0])
+            factorial2(np.array([0.0, 1.0, 2.0, 3.0]), exact=False), np.array([1.0, 1.0, 2.0, 3.0])
         )
 
     def test_special_cases_exact(self):
@@ -40,5 +38,6 @@ class TestFactorial2(unittest.TestCase):
         np.testing.assert_almost_equal(factorial2(-1.0, exact=False), 1.0)
         np.testing.assert_almost_equal(factorial2(-2.0, exact=False), 0.0)
 
-if __name__ == '__main__':
+
+if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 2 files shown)


tovrstra:
Thanks for the pull request. I'll review it and plan to merge it after #315 has been merged.

iodata/test/test_factorial2.py
```diff
@@ -0,0 +1,43 @@
+import unittest
+import numpy as np
+from .. import overlap
+
+factorial2 = overlap.factorial2
```
tovrstra, line 5: Please simplify to:
```suggestion
from ..overlap import factorial2
```

iodata/test/test_factorial2.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class TestFactorial2(unittest.TestCase):
```
tovrstra, line 8: In IOData, we don't use the `unittest` module, nor test classes, nor `self.assert...` methods. Instead, we use Pytest, which is designed to be fully functional with simple test functions. It makes writing and extending tests a lot easier. One of the nice features of Pytest is that it can deduce informative error messages from normal `assert` statements.

For testing numerical data, I'd recommend `pytest.approx`. See:

- https://stackoverflow.com/questions/8560131/pytest-assert-almost-equal
- https://docs.pytest.org/en/latest/reference/reference.html#pytest.approx

Our current code is using `numpy.testing` instead of `pytest.approx`, because a lot of that code predates `pytest.approx`. We should migrate eventually for the sake of robustness. Feel free to use `pytest.approx` already.

iodata/test/test_factorial2.py
```diff
@@ -0,0 +1,43 @@
...
+    def test_integer_arguments(self):
+        self.assertEqual(factorial2(0, exact=True), 1)
+        self.assertEqual(factorial2(1, exact=True), 1)
+        self.assertEqual(factorial2(2, exact=True), 2)
+        self.assertEqual(factorial2(3, exact=True), 3)
+        self.assertEqual(factorial2(-1, exact=True), 1.0)
```
tovrstra, line 14: Should the return value be an integer here?

tovrstra (commented):
Again, thank you for this contribution! I've added a few quick comments. I'll check in more detail before merging.

iodata/test/test_factorial2.py
D-TheProgrammer: We have followed all your comments and changed the unittest to pytest_approx. We apologize for the mistake we were basing our work on your previous comment on the issue #308, which led us to believe  that we should use unittest. The test_factorial2 has been updated and posted in this pull request. If you don't see it, please let us know, and we will create a new pull request.

Thank you, and please let us know if you merge it.

D-TheProgrammer:
@tovrstra  We have followed all your comments and changed the unittest to pytest_approx. We apologize for the mistake; we were basing our work on your previous comment on issue #308, which led us to believe we should use unittest. The test_factorial2 has been updated and posted in this pull request. If you don't see it, please let us know, and we will create a new pull request.

Thank you, and please let us know if you merge it.

tovrstra:
I just merged #315 and well merge your pull request into the current main branch to enable the Ruff linters on this pull request.

tovrstra:
It seems I cannot push the merge with the current main into your pull request. This may be related to the fact that you made a pull request from the main branch of your repository, but I'm not sure. (See branch protection rules in the GitHub documentation.) It is in general not recommended to develop on the main branch of a fork, because that makes it impossible to sync your fork with the original repository. You may have to remove and recreate your fork to fix this.

Could you create a new pull request starting from the current main? The current main has a series of new linters enabled, which prevent many anti-patterns from being committed.

When doing so, could you run the following in your local repository before making a commit?

```bash
pre-commit install
```

That way, you run the linters before committing to your git branch.

Thank you!

PR #317, opened 2024-05-29
Add PyPI release workflow

tovrstra:
This is another step to address #313.

The included workflow will make releases on PyPI and test releases on test.pypi.org.

I'm using an updated version of the trusted publisher workflow from: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
(This documentation has some outdated versions of the relevant GitHub actions, which was easy to fix.)

We will only properly see this in action (and validate that it works) after the pull request is merged into main, because that is the branch from which releases are made.


tovrstra:
I will YOLO-merge this on Wednesday June 5, unless approved earlier. Shortly after, I'll make another alpha release to make sure it all works as expected.

PaulWAyers (approved):
LGTM

tovrstra:
Thanks for the review!

PR #318, opened 2024-05-29
Migrate from attr to attrs

tovrstra:
This is another part of the renovation items in issue #313.

We used the `attr` library, which has been redesigned into the `attrs` library with an improved API. The main improvements in `attrs` are clearer names and more convenient default arguments. This is most notable in the changes of:

```python
@attr.s(auto_attribs=True, slots=True, on_setattr=[attr.setters.validate, attr.setters.convert])
```

into

```python
@attrs.define
```

This update also comes with two strongly related cleanups:

- Type hinting of NumPy arrays should be done with `numpy.typing.NDArray` instead of `numpy.ndarray`. This is a technical detail but useful because class attributes (defined with `attrs`) often have type hinting.
- Implicit optional attributes of the `IOData` object are turned into explicit optional ones, to be more in line with [PEP 484](https://peps.python.org/pep-0484/). The required changes were at the same lines that were affected by the attrs update.

I'm not sure yet if the tests will keep working on Python 3.9 because type hinting has evolved strongly over the last five Python releases. If all works, we may replace a few older type hints by more modern forms:

- `Otional[X]` -> `x | None`
- `Union[A, B]` -> `A | B`

### DIFF
--- a/iodata/basis.py
+++ b/iodata/basis.py
@@ -28,8 +28,9 @@ from functools import wraps
 from numbers import Integral
 from typing import Union
 
-import attr
+import attrs
 import numpy as np
+from numpy.typing import NDArray
 
 from .attrutils import validate_shape
 
@@ -100,7 +101,7 @@ def angmom_its(angmom: Union[int, list[int]]) -> Union[str, list[str]]:
     return ANGMOM_CHARS[angmom]
 
 
-@attr.s(auto_attribs=True, slots=True, on_setattr=[attr.setters.validate, attr.setters.convert])
+@attrs.define
 class Shell:
     """A shell in a molecular basis representing (generalized) contractions with the same exponents.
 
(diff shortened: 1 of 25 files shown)


tovrstra:
For Python 3.9, the older style of type hinting is required, `Optional[X]` instead of `X | None`, which is now fixed.

I will YOLO-merge this on Wednesday, June 5, unless approved earlier.

PaulWAyers (commented):
I'm supportive of the change, but recognize that, once again, we are in danger of diverging between QC-Devs packages, which makes things annoying. I'm OK to use `iodata` as a trial balloon for these changes, but we need a timeframe to roll (selected or all) changes out across repositories too.

tovrstra:
I understand, I'm working my way through a lot of changes these days.

I didn't mention this before, but it may be reassuring: most of these changes are based on what is or was outdated in IOData, and what I see in other (popular) Python or other repositories, the tools they use etc. I'm hoping to move to somewhat more long-lived CI solutions. Of course, you never know for sure, but if something is popular or widely endorsed, it seems worth a try.

PR #319, opened 2024-05-30
Fix factorial2 and pytest fix #308 and #309

### LINKED ISSUE #308: Fix Factorial2

D-TheProgrammer:
This pull request represents another attempt to address the issue. The previous attempt was unmergeable due to being made from the main branch. The objective of this fix is to address the issues with the `factorial2` function in the `òverlap.py`file  and create a unit test using pytest named `test_factorial2`. Our solution for the function has been validated by passing the linting checks.


D-TheProgrammer:
@tovrstra Normally it should be good now for the merge let us know 
@ludoRNLT here's the new one i forgot to tag you on this one

tovrstra (commented):
@D-TheProgrammer Thanks for the pull request and my apologies for the large number of changes we're going through while you were making pull requests.

There are still a few linters complaining. If you click on "Details" on the "pre-commit.ci - pr" line, and then scroll down, you will see the errors. Normally, when you install pre-commit in the clone of the repository on your computer, it will not allow you to make the commit when there are such issues. If you need help getting that to work, please let me know. (This is the recommended procedure, because fixing issues is always easier right after having written the code.)

If you need help with fixing the linter complaints themselves, just let me know.

iodata/overlap.py
```diff
@@ -25,32 +25,56 @@
...
-    # so np.array(n) is passed to make sure the output is always an array.
-    out = scipy.special.factorial2(np.array(n), exact=exact)
-    out[out <= 0] = 1.0
-    out[out <= -2] = 0.0
-    return out
-
-
+    # Handle integer inputs
+    if isinstance(n, (int, np.integer)):
+        if n in {-1, 0}:
```
tovrstra, line 48: Creating sets, like `{-1, 0}`, is expensive in Python
```suggestion
        if n in [-1, 0]:
```

iodata/overlap.py
```diff
@@ -25,32 +25,56 @@
...
+            return 1.0
+        if n < -1:
+            return 0.0
+        return scipy.special.factorial2(n, exact=exact)
+
+    # Handle float inputs
+    if isinstance(n, float):
+        if n in {-1.0, 0.0}:
```
tovrstra, line 56:
```suggestion
        if n in [-1.0, 0.0]:
```

iodata/overlap.py
```diff
@@ -25,32 +25,56 @@
...
+            return 1.0
+        if n < -1.0:
+            return 0.0
+        return scipy.special.factorial2(int(n), exact=exact)
+
+    # Handle array inputs
+    if isinstance(n, np.ndarray):
+        result = np.zeros_like(n, dtype=float)
+        for i, val in np.ndenumerate(n):
+            if val in {-1.0, 0.0}:
```
tovrstra, line 66:
```suggestion
            if val in [-1.0, 0.0]:
```

D-TheProgrammer:
> @D-TheProgrammer Thanks for the pull request and my apologies for the large number of changes we're going through while you were making pull requests.
> 
> There are still a few linters complaining. If you click on "Details" on the "pre-commit.ci - pr" line, and then scroll down, you will see the errors. Normally, when you install pre-commit in the clone of the repository on your computer, it will not allow you to make the commit when there are such issues. If you need help getting that to work, please let me know. (This is the recommended procedure, because fixing issues is always easier right after having written the code.)
> 
> If you need help with fixing the linter complaints themselves, just let me know.


We tried with our computers and we had 4 or 5 linters that didn't work but they were not tied to our fix it was about the other functions outside of factorial2 but still in overlap.py, this is why we did our PR because we thought that someone else was working on those other issues. 
We

tovrstra:
I have another small question. I may be overlooking something, but I don't believe we need the floating point case in IOData and always need exact=True, which may be used to simplify our version of `factorial2` considerably.

tovrstra:
> @tovrstra We tried with our computers and we had 4 or 5 linters that didn't work but they were not tied to our fix it was about the other functions outside of factorial2 but still in overlap.py, this is why we did our PR because we thought that someone else was working on those other issues. 

Ok, sounds good. No worries. Thanks for the quick reply. I'll take care of it after the weekend

tovrstra:
@D-TheProgrammer I made all improvements I could think of. Can you double-check the final result? Thanks.

tovrstra:
Thanks for checking. I'll merge soon.

PR #320, opened 2024-06-03
Use sys.executable instead of "python".

tovrstra:
This is a silly PR meant to test some of the newly activated code analysis tools. Despite being minor, it still is useful.

I'll YOLO-merge this in a few days if all tests pass.

### DIFF
--- a/iodata/test/test_cli.py
+++ b/iodata/test/test_cli.py
@@ -21,6 +21,7 @@
 import functools
 import os
 import subprocess
+import sys
 
 from numpy.testing import assert_allclose, assert_equal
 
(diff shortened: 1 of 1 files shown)


PaulWAyers (approved):
Again, OK, but I worry about `iodata`'s workflow/packaging/dependencies/setup diverging from other repositories. It risks making `iodata` hard to maintain if it uses tools that none of the other repositories use. But OK to try things out for one repository, though perhaps a repo that is more actively being developed would be a better choice for testing.

tovrstra:
The automatic pull request review is definitely an experiment at this stage.

I haven't used them enough to be able to judge their utility in general. So far, they did help to find some ugly pieces of code that were somehow overlooked. Compared to the other changes I'm making, setting up deepsource and others, is quite low effort.

tovrstra:
Thanks for reviewing!

PR #321, opened 2024-06-04
Fix documentation build issues

tovrstra:
Our documentation build on read-the-docs (RTD) is failing. This should fix it, hopefully.

Because this can only be fully tested by merging the PR and monitoring the doc build in RTD, I will merge this soon.

Ideally, we get rid of RTD for the following reasons:

- RTD supports only one admin account per project, i.e. bus-factor 1 by construction, unless one shares passwords, which is a bad idea anyway.
- Testing the RTD build is only possible after merging.

As an alternative, we can build the documentation in a GitHub Action and host it on gh-pages instead.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,28 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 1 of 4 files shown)


PaulWAyers (approved):
I agree, let's ditch read-the-docs. We're mostly using GitHub pages for everything else now anyway, and it's much nicer to deal with.

PR #322, opened 2024-06-04
Provide some guidance for keeping the Changelog up-to-date

tovrstra:
This is one of the smaller items on the list in #313.

For more details, see https://keepachangelog.com/en/1.1.0/ (It is written for Markdown, but easy enough to follow the same practices in restructured text.) This is not a big change, mainly providing some more guidance on how to write helpful changelogs.

This is mainly going to be of use when we have a first stable release, but it is better to have it in place already before that time.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -20,11 +20,19 @@
     :
     : --
 
-IOData Changelog
-=================
+Changelog
+#########
 
+All notable changes to this project will be documented in this file.
 
-**Version 1.0.0**
+The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.1.0/>`_,
+and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`_.
+
+`Unreleased`_
+=============
+
+Changed
+-------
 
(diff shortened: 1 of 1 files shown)


tovrstra:
I'll YOLO-merge this on Wednesday, June 12, unless reviewed earlier.

PaulWAyers (approved):
Sure, seems like a good idea.

I'm curious (unless I'm missing something, this is a manual changelog format) whether we can do something automated. E.g., can we have a template for a pull request that ends up automatically updating the change log? 

Probably not worth the effort to automate, though, especiall for a relatively mature packages like `iodata`.

tovrstra:
Thanks for the comments.

I know there are scripts to extract the changelog from the commit history, but using these correctly is hard. It is not only a matter of creating the scripts that do such work. That biggest hurdle is to get everyone to format their commit messages or pull requests, such that the changelog looks good in the end. The guideline I'm linking to advices against such tools. I haven't really tried this myself.

With AI, it may be possible to derive nice changelogs from the Git history, but that seems a little too crazy right now. Who knows, one day ...

PR #323, opened 2024-06-04
Remove import_resources try-except block

tovrstra:
Another small item on the list of #313 

These weird imports were only relevant for Python versions that are past their end of life: 3.6 and older.

## Summary by Sourcery

This pull request removes the try-except blocks used for importing 'importlib_resources' to support older Python versions (3.6 and older). The code now directly imports from 'importlib.resources', simplifying the import statements and removing compatibility code for unsupported Python versions.

* **Enhancements**:
    - Removed try-except blocks for importing 'importlib_resources' in favor of direct imports from 'importlib.resources'.

PR #325, opened 2024-06-04
Update installation instructions

tovrstra:
Another step in #313 (before the next prerelease. I'm aiming for Wednesday, June 12)

These instructions are intended to work with the next prerelease:
- I'm not preparing conda releases at this stage. (Not so useful anymore now that IOData is pure Python. Pip works in a conda env.)
- I've updated NumPy and SciPy dependencies to the oldest ones that are not listed on the NIST national vulnerability database. It may be worth extending our CI to explicitly test with these verions.
- The `--pre` option is no longer needed.
- The number installation options has been reduced:
  - No conda releases.
  - We no longer have setup.py (because such setup scripts are no longer recommended).
  - The option `--user` is no longer needed. `pip` assumes it when there is no venv.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,37 +50,14 @@ Please use the following citation in any publication using IOData library:
 Installation
 ------------
 
-..
-    : To install IOData using the conda package management system, install
-    : `miniconda <https://conda.io/miniconda.html>`__ or
-    : `anaconda <https://www.anaconda.com/download>`__ first, and then:
-    :
-    : .. code-block:: bash
-    :
-    :     # Create a horton3 conda environment. (optional, recommended)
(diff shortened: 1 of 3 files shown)


tovrstra:
P.S. Our CI is already pushing test releases to test.pypi.org, if you want to get a preview. https://test.pypi.org/project/qc-iodata/#history

PaulWAyers (approved):
Thanks Toon.

tovrstra:
I'll wait with merging this one, so I can merge it shortly after #322 and #323, and follow up closely after by making a prerelease. (It would be good to have all of them in the pre-release.)

PR #331, opened 2024-06-08
Stricter testing

tovrstra:
- When testing with the old Python version (3.9 at the moment), the oldest supported versions of the dependencies are installed.
- The oldest supported SciPy is increased from 1.8.0 to 1.11.0, due to required array support in factorial2.
- The option `-W error` is added to pytest, so that it fails in case of a warning.

This is part of the changes listed in #313. The first bullet was brought up in #325.

I will YOLO-merge this on Friday, June 14 unless reviewed earlier.

## Summary by Sourcery

This pull request enhances the CI workflow by ensuring that the oldest supported versions of dependencies are installed when testing with Python 3.9. Additionally, it configures pytest to treat warnings as errors, improving the strictness of the tests.

* **CI**:
    - Added a step to install the oldest supported versions of dependencies when testing with Python 3.9.
    - Included the `-W error` option in pytest to treat warnings as errors.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -24,7 +24,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install oldest versions of supported dependencies
         if: ${{ matrix.python-version == "3.9"}}
-        # Changes to these dependencies must be updated
+        # Ensure changes to these dependencies are reflected
         # in pyproject.toml and docs/install.rst
         run: pip install numpy==1.22 scipy==1.8 attrs==21.3.0
       - name: Install development version
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -26,7 +26,7 @@ jobs:
         if: ${{ matrix.python-version == '3.9'}}
         # Ensure changes to these dependencies are reflected
         # in pyproject.toml and docs/install.rst
-        run: pip install numpy==1.22 scipy==1.8 attrs==21.3.0
+        run: pip install numpy==1.22 scipy==1.10 attrs==21.3.0
       - name: Install development version
         run: pip install -e .[dev]
       - name: Run Pytest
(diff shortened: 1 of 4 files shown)


PaulWAyers (commented):
Is there any way to support earlier versions of Python?

We have a *major* annoyance right now because [HORTON 2.3 supports only up to Python 3.9](https://github.com/theochem/horton/blob/764dd2fa958aee0dcee30b19ed0520d65d83b45a/conda/conda_build_config.yaml#L3). This makes it impossible to use "new HORTON" features in the same environment as HORTON 2.x, and there is stil a lot of old workflows that use facets of HORTON 2.x that haven't been migrated yet.

So I guess the two questions are, and I guess @tovrstra  and @matt-chan are the people to answer them:
1. Is it better to keep HORTON 3.x things compatible down to Python 3.9
2. Is it better to upgrade HORTON 2.x to support Python past 3.9

One of these two things needs to be done, IMHO.

PaulWAyers:
> mp 2.x compatibility past 3.9. The main blocker is getting the newer versions of the dependencies.

Can you take the lead on making this update to HORTON 2.3 then (preferably sooner rather than later)? I'll then approve this as the point will be moot by the time it's relevant.

PaulWAyers (approved):
@matt-chan reassured me that we can bump HORTON 2.3 to Python 3.11 or later in short order, so it should be fine.

tovrstra:
Sounds good. Thanks for reviewing and checking all this.

IOData works for Python 3.9 now and we can easily keep it that way while it is officially supported, i.e. until October 2025. (See https://devguide.python.org/versions/)

The reason I did not consider supporting 3.8 is that it goes end of life soon, i.e. October 2024.

Note that Python 3.9 is also the oldest version that supports [type hinting with built-in types (PEP 585)](https://peps.python.org/pep-0585/) without `__future__` imports, which is nice to have. It makes the source code easier to write and read.

PR #332, opened 2024-06-12
Update contributing guide

tovrstra:
This is another step in #313. 

I've revised the document to be up-to-date with the recent changes listed in #313 and tried to improve the readability. 

Some points were taken from the central [contributing guide](https://github.com/theochem/guidelines/blob/main/contributing.md) and updated. Eventually, most of it can be centralized (the first four sections). There is also a difference in scope: this document is a bit more tutorial-like than the central one.

Direct link to the new version (easier to read than the diff): https://github.com/tovrstra/iodata/blob/1032c13485a2dba0f8298617028145d806d5dc1d/CONTRIBUTING.rst

## Summary by Sourcery

This pull request updates the contributing guide to reflect recent changes and improve readability. It also updates pre-commit hooks to newer versions and removes an unnecessary noqa comment in the code.

* **Enhancements**:
    - Updated pre-commit hooks to newer versions in .pre-commit-config.yaml.
* **Documentation**:
    - Revised the contributing guide to be up-to-date with recent changes and improved readability.
* **Chores**:
    - Removed unnecessary noqa comment in iodata/utils.py.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,11 +27,11 @@ repos:
   hooks:
     - id: remove-crlf
 - repo: https://github.com/python-jsonschema/check-jsonschema
-  rev: 0.28.4
+  rev: 0.28.5
   hooks:
     - id: check-github-workflows
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.4.5
+  rev: v0.4.8
   hooks:
     - id: ruff-format
     - id: ruff
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,50 +1,260 @@
-We'd love you to contribute. Here are some practical hints to help out.
+We'd love you to contribute.
+Here are some practical tips to help you get started.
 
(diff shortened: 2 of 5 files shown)


PaulWAyers (approved):
It looks good. I created an [issue in `theochem/Guidelines`](https://github.com/theochem/guidelines/issues/2) so we don't lose track of the need to harmonize this across repositories (though the last part of this document is specific to `IOData`.

tovrstra:
@PaulWAyers Thanks for reviewing. 

Good point. I'll write out the whole CI setup of IOData in an imminent pull request. For now, the main missing piece for IOData is coverage analysis. It is not so hard to get to work, but it makes testing so slow.

More general versions of this and the next PR are certainly desirable. When working these out, it would be good to get an idea of what everyone else's perspective on the matter is. For a given selection of tools, I can cook up the most recent way to use them, but which tools to use remains (somewhat) a matter of taste. There are a few different tools one can use for each CI task, and they all have their strengths. A bit of diversity is actually also useful to gather more experience.

Issue #254, opened 2021-04-01
Add separate, more detailed section for the Input Writer?


wilhadams:
Adding things like non-standard basis set options (#221) are making it possible to do a lot of useful and complex things with the input writer, which is really exciting. However, when I was previously adding examples for using custom templates, I started to worry that the "Input Files" section of the getting started page was beginning to get pretty lengthy. Would it make sense to break out most of the input writer details into a new page, and leave just the basic usage in the getting started page? 

@tovrstra @FarnazH @PaulWAyers

tovrstra:
Yes, most definitely. It would be good to have a separate section with more elaborate examples, while keeping the "Getting Started" is kept short and simple.

While doing this, it could be good to address #210 too. Executing the more elaborate examples int the unit tests would guarantee the examples keep working. (This also implies the necessary input files are present to run the examples, which would also be helpful.)

Issue #210, opened 2020-09-01
Extend the examples in `getting_started.rst` and run them as unit tests


tovrstra:
The examples in `getting_started.rst` are a bit limited and could be extended (in line with the manuscript). They should also be tested in the unit tests to make sure they keep working. This can be accomplished by writing them as python scripts in the `iodata/test/data` directory. These files can be easily accessed by the unit tests and included in the documentation with the [literalinclude directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude).

tovrstra:
Adding more examples is not critical for 1.0.0, so I'm removing this issue from the milestone.

PR #351, opened 2024-06-23
Split the getting started docs

tovrstra:
This fixes one of the tasks in #191 and is also useful for #254 and #210.

The getting started guide is split in smaller bite-size parts. This also creates some more room for extending the getting started with a few more examples as mentioned in some of the cited issues.

The contents of the guide has not really changed, except maybe for the text in `gettingstarted/index.rst`

I will YOLO-merge this on Friday, June 28, unless reviewed earlier.


PaulWAyers (approved):
Thanks Toon. I especially appreciate the way the units is explained clearly, with the rationale of using old-timey notation.

tovrstra:
Thanks for checking. Merging...

Issue #204, opened 2020-08-27
Janitoring todo list before 1.0.0


tovrstra:
I'll keep a list of minor cleanups that should be done. It is easier to carry these out when all PRs are merged because these cleanups touch many parts of the code, easily causing merge conflicts:

- [ ] Check `# pylint: disable=` scopes. I noticed I've been placing them incorrectly, see e.g. this fix: https://github.com/theochem/iodata/pull/196/commits/7ea93001fe44182b4d8752588e84e26f49c66cfc

PR #354, opened 2024-06-28
Make __all__ immutable

tovrstra:
This is one on the list in #204, just a minor cleanup, to avoid making such changes *along the way* in future pull requests.

Making __all__ a tuple is not a tradition in Python, but makes sense for things that are not supposed to be variable.


PaulWAyers (approved):
OK by me.

tovrstra:
Thanks, merging...

PR #355, opened 2024-06-28
Require optional arguments of iodata.api functions to be specified with keywords

tovrstra:
This fixes one of the tasks in #191, deferred from #352.

This way, scripts using IOData become more readable, lower risk of bugs. It also gives us the freedom to reorganize optional arguments with less risk of breaking API compatibility. This change is in itself API-breaking.


PaulWAyers (approved):
LGTM

tovrstra:
Thanks for checking! Mergin...

PR #356, opened 2024-06-28
Do not use iodata as variable name

tovrstra:
Fixes another task on the list of #191.

Having `iodata` as a variable name is bad because this is also the package name. This fixes it in a way that is consistent with the rest of the IOData code, e.g. as in the file format modules.


PaulWAyers (approved):
Fine by me.

tovrstra:
Thanks for checking! Going in...

Issue #256, opened 2021-04-03
Clarifications and documentation improvements in `iodata.basis`


tovrstra:
(This was a part of #146.)

There are some points of confusion in the documentation of `iodata.basis` and some names may be confusing. I'll maintain a list of suggestions for improvements below.

- [x] `Shell.nprim`: The "number of primitives" can be interpreted in a few different ways. E.g. `nexp` for number of exponents has less ambiguity. `contraction_length` is too long. See #357
- [ ] `Shell.ncon`: It should be clarified that different contractions within one generalized contraction may, but not must, have different angular momenta. They share the same exponents but have different contraction coefficients. See #359
- [ ] `MolecularBasis.get_segmented`. The terminology of loop "unrolling" in the docstring is confusing. See #373

(Suggestions are welcome.)

Some of the points above imply minor API changes.

PR #357, opened 2024-07-04
Rename `nprim` to `nexp` where appropriate

tovrstra:
This is an attempt to solve a terminology issue that has caused a lot of confusion in the past. It addresses part of #256. (I'm going through these clarifications before continuing with #191 to make future PRs related to #191 easier to process.)

The term *number of primitives*, abbreviated now in IOData is `nprim` (property of the `Shell` class), has different interpretations in different context and file formats:

- In the FCHK format, the field `Number of primitives per shell` is an array with contraction lengths. Because this is one of the earliest supported file formats, it also led to the name `nprim` in the `Shell` class.
- In the WFN and WFX formats, the `GAUSSIAN` line contains a number of primitives, e.g. `74 PRIMITIVES`, which refers to the total number of individual basis functions, e.g. three primitives for a p-type shell (X, Y and Z), etc.
- In the MWFN format, `Nprims` refers to the sum of all contraction lengths, counting those of SP generalized contractions twice, and `Nprimshell` is the sum of all contraction lengths, counting those of SP generalized contractions once.

I've considered a few alternatives for the name of the `Shell.nprim` property:

- I do not like `contraction_length`, too long. This name is also assumes some familiarity with basis set terminology.
- `nexp` (as in this PR) refers to its literal meaning, i.e. *number of exponents* in a contracted shell. (It is also implemented as `return len(self.exponents)` I prefer this one because it is short and semantic. It also uses the `nsomething` form to refer to the size of an array along one of its dimensions, as found elsewhere in IOData.
- Also `nexponent` is possible, a bit clearer, more typing.

I'm trying to make only one name change per PR, to keep things easy to review. In case of concerns about other attribute names, properties, please add them to #256.

`nprim`-like variables in the WFN, WFX and MWFN formats were not renamed when they referred to the meaning of that file format (and not the one we had in the Shell class.)

I will YOLO-merge this on Wednesday, July 10 unless reviewed earlier.


tovrstra:
@sourcery-ai review

PaulWAyers (approved):
Thanks for the detailed explanation. I think nexp is a clear name, and the documentation seems clear when it is invoked too.

tovrstra:
Thanks for reviewing Paul!

PR #376, opened 2024-12-17
Fix unrecognized Gaussian input format: *.com and *.gjf

FanwangM:
The current implementation does not support .com and *gjf because it used module name instead of "PATTERNS".

## Summary by Sourcery

Bug Fixes:
- Fix support for recognizing Gaussian input formats with extensions .com and .gjf by using patterns instead of module names.


PaulWAyers (commented):
Looks OK to me but I'd like an opinion from someone who does a lot more work with IOData.

FanwangM:
I figured out the right way to use the database, and this should not be seen as a bug. So I am closing it here.

PR #378, opened 2025-03-14
Fchk ref energy

### LINKED ISSUE #377: Problem when reading the total energies of a Gaussian16 optimization fchk
Hi all,

I was reading the energies from a simple water optimization with Gaussian16 using:
```
[f.energy for f in load_many("out.fchk", fmt="fchk")]
```
I noticed, however, that I do not get the correct results since in the fchk the line
```
Opt point       1 Results for each geome   R   N=           6
```
Is preceded by
```
Optimization Reference Energy              R     -7.600000000000000E+01
```
So the energy that iodata reads when looking for `1 Results for each geome` (in `iodata.formats.fchk.py`) is actually not the total energy but the difference with this reference energy. I am not sure if/when this was introduced in Gaussian16, I have noticed it is present in both revision B and C but I do not have information on older version. I could also not really find any information online.

The full fchk is attached. Do you think it'd be possible to look for this reference energy and add it to the ones from the results (or, of course, let me know if I have overlooked something)?

Thanks in advance,
Massimo

[input.fchk.txt](https://github.com/user-attachments/files/19128570/input.fchk.txt)

tovrstra:
Fixes #377 

The first commit is the actual fix.

The second commit was needed to make things work with a recent version of pre-commit. Some of the versions in .pre-commit-config.yaml were no longer compatible. All have been updated to the latest versions, and these new versions fix more code smells.


tovrstra:
@mbocus Can you test the version from this PR locally to verify that it also works for you? You should be able to install it with:

```bash
pip install git+https://github.com/theochem/iodata.git@refs/pull/378/merge
```

mbocus:
@tovrstra Many thanks for the swift fix! I've run some tests and, from what I can see, everything seems to be working as expected.

tovrstra:
Thanks foro the feedback!

PaulWAyers (approved):
Looks fine to me. Thanks @tovrstra

tovrstra:
Thanks for checking!

PR #380, opened 2025-05-30
Use Python 3.13 as newest Python version for pytest workflow

tovrstra:
This would make future releases formally installable in a Python 3.13 environment.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.9', '3.12']
+        python-version: ['3.9', '3.13']
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


FanwangM (approved):
The changes look great to me. One minor point: do we want to test against all Python versions for Python >= 3.9 instead of `['3.9', '3.13']` ? This might lead to longer test durations, but more structured testing guarantees our package functions effectively across all versions.

@tovrstra

tovrstra:
I think it makes sense to test on more Python version, but does indeed become slower. We might also consider testing on all Python version on Linux only. Similarly, we can test on all OSes with just one Python version? (Testing is notably slower on macOS virtual machines.)

FanwangM:
This proposal looks great. How about this? We only test Python 3.13 for MacOS and Windows, but all Python version for Ubuntu.

@tovrstra 
```Python
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
        exclude:
          # only test Python 3.9 and 3.13 for MacOS and Windows
          - os: macos-latest
            python-version: ['3.9', '3.10', '3.11', '3.12']
          - os: windows-latest
            python-version: ['3.9', '3.10', '3.11', '3.12']
```

PR #381, opened 2025-05-31
Test updated schema for adding Python 3.13 support.

FanwangM:
## Summary by Sourcery

Add Python 3.13 support by updating CI test matrix and project metadata

Enhancements:
- Add Python 3.13 to the project's classifiers in pyproject.toml

CI:
- Include Python 3.13 in the GitHub Actions pytest matrix and add Python 3.10 and 3.11 to the strategy
- Restrict macOS and Windows runs to only Python 3.9 and 3.13

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -19,9 +19,9 @@ jobs:
         exclude:
           # only test Python 3.9 and 3.13 for MacOS and Windows
           - os: macos-latest
-            python-version: ['3.10', '3.11', '3.12']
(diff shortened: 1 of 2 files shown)


tovrstra (commented):
I believe exclusion need to be listed individually, e.g. as follows:

```yaml
exclude:
  - os: macos-latest
    python-version: '3.10'
  - os: macos-latest
    python-version: '3.11'
...
```

It's a bit tedious, but AFAIK there is no other way. The only alternative I can think of is to define the matrix with Linux only and then to use `include:` to add a few more OS-Python combinations

Issue #386, opened 2025-10-21
load_many from fchk fails if the Gaussian job is an optimization of a single atom


mbocus:
Hi all, 

I have stumbled upon an issue when loading a formatted checkpoint file of Gaussian (*.fchk). I have a script that checks the job type and, if it is an optimization, uses the `load_many` function to parse the output. If the optimization of a species with more than one atom finishes in one step (i.e., the structure is already optimized), all works as intended. For example:
```
for mol in load_many("water.fchk"):
    print(mol.atcoords)
```
returns
```
[[-3.13242678e-34 -7.31649715e-35  2.40261291e-01]
 [-2.33989326e-32  1.43244075e+00 -9.61045165e-01]
 [-1.75423398e-16 -1.43244075e+00 -9.61045165e-01]]
```
On the other hand, in the limiting case in which a single atom is "optimized" (which is obviously meaningless, but something that can happen in automated workflows), the reading fails with the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mbocus/miniconda3/envs/iodata_env/lib/python3.11/site-packages/iodata/api.py", line 219, in load_many
    for data in format_module.load_many(lit, **kwargs):
  File "/home/mbocus/miniconda3/envs/iodata_env/lib/python3.11/site-packages/iodata/formats/fchk.py", line 329, in load_many
    raise LoadError("Cannot find IRC or Optimization trajectory in FCHK file.", lit)
iodata.utils.LoadError: Cannot find IRC or Optimization trajectory in FCHK file. (atom.fchk:1044)
```
I am using the latest iodata release (v1.0.0a8) and, attached, you may find the two fchk files I used for testing.

Thanks a lot in advance,
Massimo

[water.fchk.txt](https://github.com/user-attachments/files/23015925/water.fchk.txt)
[atom.fchk.txt](https://github.com/user-attachments/files/23015926/atom.fchk.txt)

Ao-chuba:
Hi, @tovrstra Gaussian omits the Optimization Number of geometries block for single step optimizations, even though the file is still marked as an optimization run (FOpt). This causes load_many to fail early.
If the trajectory block lookup fails (LoadError), we shouldn’t give up immediately. Instead, we check whether the file is marked as an optimization run. If it is, we can safely treat it as a trajectory of length 1 and construct a single frame using the standard Current Cartesian Coordinates, Total Energy, and Cartesian Gradient fields, which are already present.

[...]

PR #385, opened 2025-10-09
298/ Fixed naming conventions

dhruvDev23:
changed the variable names based on the convention
attypes => atnames
attype => atname
restypes => resnames
restype => resname

## Summary by Sourcery

Enhancements:
- Replace attypes/restypes with atnames/resnames and attype/restype with atname/resname in load_one and dump_one functions

PR #387, opened 2025-10-22
Drop support for Py 3.9

tovrstra:
This is an attempt to fix the failing Sphinx build as seen in #385. The failing build is related to lacking support for the EOL Python 3.9. (See https://devguide.python.org/versions/ for the timeline.)

@mbocus: I'm going to deal with this one first to get rid of some old cruft and get the CI running again. I'll pick up #386 next. I assume you're not using Py3.9 for the course?

@PaulWAyers: Do you expect ending 3.9 support to be an issue? Is anyone in your group critically depending on it? Because this is a dead Python version, maintaining compatibility with it causes compatibility issues with other packages that already moved on, in this case Sphinx and related packages.

## Summary by Sourcery

Remove Python 3.9 support and modernize the codebase to leverage Python 3.10+ type features

Enhancements:
- Replace typing.Optional and typing.Union annotations with PEP 604 union syntax and built-in generics throughout the codebase
- Add strict=True to zip calls in various modules for stricter argument matching

Build:
- Bump requires-python to 3.10 in pyproject.toml
- Upgrade pre-commit hook revisions (pre-commit-hooks, check-jsonschema, ruff)

CI:
- Drop Python 3.9 from GitHub Actions test matrix and use Python 3.10–3.13 across platforms
- Set Sphinx build workflow to run on Python 3.12 only

### DIFF
--- a/tools/harmonics.py
+++ b/tools/harmonics.py
@@ -86,7 +86,7 @@ def get_bare_transforms(ellmax: int):
             poly = sp.Poly(rrsh.pop(0), x, y, z)
             lookup = {
                 (int(nx), int(ny), int(nz)): coeff
-                for (nx, ny, nz), coeff in zip(poly.monoms(), poly.coeffs(), strict=False)
+                for (nx, ny, nz), coeff in zip(poly.monoms(), poly.coeffs(), strict=True)
             }
             for nx, ny, nz in iter_mononomials(ell):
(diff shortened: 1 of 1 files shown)


tovrstra:
Since there are no immediate expressions of concern regarding old Python versions, I'll merge this one.

PR #390, opened 2025-11-26
Update for load_one info in gro files

lmacaya:
This PR is a small update for gro files without velocities data. Also, I added a dictionary in the periodic.py file for atomic mass and element symbol relation (atmass2sym).

## Summary by Sourcery

Handle GRO files that omit velocity columns and expose atom names in GROMACS loader metadata while adding element–mass lookup tables.

New Features:
- Introduce element atomic-mass-to-symbol and symbol-to-atomic-mass mapping dictionaries in the periodic data module.

Bug Fixes:
- Allow loading GRO trajectory frames that do not include velocity columns by treating velocities as optional and propagating None when absent.
- Expose atom names rather than atom types in the force-field parameter metadata returned by the GROMACS loader.

Tests:
- Add regression test and sample GRO file without velocities to verify correct handling of missing velocity data and metadata.


lmacaya:
Hi @tovrstra! I'm using gromacs gro and top files to populate an input file. Gro files does not have element info, so i use the atomic masses from the top file to guess every element. I was checking the AtomDB package and it has all this information. Based on this, it's not necessary to include it in IOData, just to avoid redundancy.

tovrstra (approved):
Thanks for the improvement. I checked once more and ran the tests also locally. All looks good.

